PR #1, opened 2015-06-24
Generalize RPM version comparision

cawallin:
Change PRESTO_RPM_BASENAME and PRESTO_VERSION to be compared via a regex,
since we don't want to hard code the version of the RPM

Still to come in the next patch: not hard coding the RPM filename itself,
but that is a bit involved so it's better for it to be in a separate patch.

Task: SWARM-811
Testing: make clean test-all with new RPM


cawallin:
@rschlussel @anusudarsan review, please

rschlussel:
lgtm

PR #2, opened 2015-06-24
Detect RPM in prestoadmin main dir, else download

cawallin:
* detect if Presto RPM is in the prestoadmin main dir
* otherwise, download from s3 if necessary
* remove all instances of PRESTO_RPM

Task: SWARM-811
Testing: make clean-all test-all

### DIFF
--- a/tests/product/base_product_case.py
+++ b/tests/product/base_product_case.py
@@ -35,10 +35,7 @@ from tests.docker_cluster import DockerCluster, DockerClusterException, \
 
 DIST_DIR = os.path.join(prestoadmin.main_dir, 'tmp/installer')
 
-# TODO: make tests not dependent on the particular version of Presto at
-# http://teradata-download.s3.amazonaws.com/aster/presto/lib/presto-0.101-1.0.x86_64.rpm
-PRESTO_RPM = 'presto-0.101-1.0.x86_64.rpm'
-PRESTO_RPM_BASENAME = r'presto-.*'
+PRESTO_RPM_GLOB = r'presto-*.x86_64.rpm'
 PRESTO_VERSION = r'presto-main:.*'
 
 
@@ -82,6 +79,21 @@ task.max-memory=1GB\n"""
         super(BaseProductTestCase, self).setUp()
         self.maxDiff = None
         self.docker_client = Client(timeout=180)
+        self.presto_rpm_filename = self.detect_presto_rpm()
+
+    def detect_presto_rpm(self):
+        """
+        Detects the Presto RPM in the main directory of presto-admin.
+        Returns the name of the RPM, if it exists, else returns None.
+        """
+        rpm_names = fnmatch.filter(os.listdir(prestoadmin.main_dir),
+                                   PRESTO_RPM_GLOB)
+        if rpm_names:
+            # Choose the last RPM name if you sort the list, since if there
+            # are multiple RPMs, the last one is probably the latest
+            return sorted(rpm_names)[-1]
+        else:
+            return None
 
     def setup_docker_cluster(self, cluster_type='centos'):
         cluster_types = ['presto', 'centos']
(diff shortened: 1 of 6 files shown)


cawallin:
please review @petroav @rschlussel

tests/product/base_product_case.py
```diff
@@ -39,7 +39,7 @@
 
 # TODO: make tests not dependent on the particular version of Presto at
 # http://teradata-download.s3.amazonaws.com/aster/presto/lib/presto-0.101-1.0.x86_64.rpm
-PRESTO_RPM = 'presto-0.101-1.0.x86_64.rpm'
+PRESTO_RPM_GLOB = r'presto-*.x86_64.rpm'
```
rschlussel: If we have the rpm filename, then instead of doing a regexp, we should use that.

tests/product/test_server_install.py
```diff
@@ -335,10 +336,11 @@ def test_install_interactive_with_ips(self):
         self.copy_presto_rpm_to_master()
 
         cmd_output = self.run_prestoadmin_script(
-            "echo -e 'root\n22\n%s\n%s\n' | "
-            "./presto-admin server install /mnt/presto-admin/%s " %
+            'echo -e \'root\n22\n%s\n%s\n\' | '
```
rschlussel: so I guess our style is to use single quotes even when it means escaping?

tests/product/base_product_case.py
cawallin: The idea is that you can put any RPM in the main directory of presto-admin and it'll work. we save the actual name below at presto_rpm_filename

tests/product/test_server_install.py
cawallin: eh, I can change it back. I was using single quotes no matter what, but I don't feel that strongly about it.

tests/product/base_product_case.py
rschlussel: Oh wait never mind.  You already do the thing that I was trying to say.

tests/product/test_server_install.py
rschlussel: no it's fine.  Just noting so we can be consistent.

rschlussel:
You should check the status tests

cawallin:
I did a test-all. The RPM version that status checks was changed in the previous commit; it just checks that the status Presto version matches presto-.*.  That seemed like the most reasonable thing, because the version that prints out in status can be super weird (now it's something like presto-0.107-ac-ade123a-dirty), and we aren't testing that it prints out the correct presto version necessarily.

rschlussel:
okay.  looks good

tests/product/base_product_case.py
```diff
@@ -402,7 +424,7 @@ def assert_started(self, process_per_host):
         return process_per_host
 
     def escape_for_regex(self, expected):
-        expected = expected.replace('{rpm}', PRESTO_RPM)
+        expected = expected.replace('{rpm}', self.presto_rpm_filename)
```
rschlussel: A few more comment (some more related to your previous patch)
1. is the basename, just the rpm name without .rpm at the end?  In that case, it should be changed also.  Also, if that's true then we don't actually need a regex here.  For most cases we should just replace the {rpm}/ {rpm_basename} and do a straight up equals or EqualsIgnoringOrder comparison.  
2. There's already a method  re.escape(). 
3. this method would be a problem if you want to replace rpm names, but have actual regexes in the line (something that is using the [] or () in a regex-y way.

cawallin: 1. There's one case where the basename isn't quite that (but I guess in that case, it's not actually the basename then). I can change it.
2. That escapes too many things and I got a regex parsing error when I used it!
3. yes. I did it like this because I didn't want to change all of the text files to escape the []() characters.

tests/product/base_product_case.py
```diff
@@ -228,21 +243,28 @@ def upload_topology(self, topology=None):
...
         self.docker_cluster.exec_cmd_on_container(
             self.docker_cluster.master, 'rpm -K --nosignature '
-            + os.path.join(DOCKER_MOUNT_POINT, PRESTO_RPM))
+            + os.path.join(DOCKER_MOUNT_POINT, self.presto_rpm_filename))
 
     def copy_presto_rpm_to_master(self):
-        rpm_path = os.path.join(prestoadmin.main_dir, PRESTO_RPM)
-        if not os.path.exists(rpm_path):
+        if not self.presto_rpm_filename:
+            # TODO: once the RPM is on Maven Central, pull the RPM from there
```
petroav: +1

tests/product/test_server_install.py
petroav: I think when quotes are nested we should alternate between double and single. For a string by itself, I've been converting everything to single quotes. They look neater to me.

petroav:
lgtm % tests passing

cawallin:
Updated -- @petroav you should probably take a look, because I changed some stuff that might affect your patches, since now there's a general replace_keywords() method called for run_prestoadmin and run_prestoadmin_script which replaces master, slave[1..3], rpm, etc. I'm willing to wait and be the one to merge conflicts, if that makes life easier for you :)

petroav:
Thanks for the heads up. You should go ahead and push. I'm not ready yet and it shouldn't be too much of a hassle to rebase on top of what you change here.

PR #3, opened 2015-06-25
Error handling: Abort current task on failure

rschlussel:
Change behavior to abort the running task for a particular host on
failure, so subsequent tasks don't get run.  Continue with tasks for
other hosts.  This means that if there is an issue with one node, other
nodes are not affected, but the task on that particular node is stopped.

Testing: make test-all

### DIFF
--- a/tests/product/resources/connector_permissions_warning.txt
+++ b/tests/product/resources/connector_permissions_warning.txt
@@ -1,24 +1,28 @@
 
-Warning: [master] Error validating /etc/opt/prestoadmin/connectors/tpch.properties
+Fatal error: [%(slave3)s] Error validating /etc/opt/prestoadmin/connectors/tpch.properties
 
 Underlying exception:
     Permission denied
 
+Aborting.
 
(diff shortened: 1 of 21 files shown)


rschlussel:
@cawallin @anusudarsan

tests/product/resources/connector_permissions_warning.txt
```diff
@@ -1,24 +1,28 @@
 
-Warning: [master] Error validating /etc/opt/prestoadmin/connectors/tpch.properties
+Fatal error: [%(slave3)s] Error validating /etc/opt/prestoadmin/connectors/tpch.properties
```
akshatnair: Any particular reason you have host names enclosed within %s or this can be avoided?

rschlussel: We had spoken about not hard coding the hostnames for the docker cluster so that you can run more than one test at a time, so I didn't want to use hardcoded names for the things I was changing.

PR #4, opened 2015-06-25
pa test infra: move functionality over to docker_cluster

petroav:
* Move methods that provide general docker functionality over to
  docker_cluster.py.
* Make the local and remote mount dirs instance variables. This prevents
  mismatches when the docker cluster is started with one mount point and then
  some other method specifies another mount point.

Task: SWARM-790

### DIFF
--- a/tests/docker_cluster.py
+++ b/tests/docker_cluster.py
@@ -27,13 +27,14 @@ from time import sleep
 
 from docker import Client
 from docker.errors import APIError
+from nose.tools import nottest
 
 from prestoadmin import main_dir
 
 INSTALLED_PRESTO_TEST_MASTER_IMAGE = 'teradatalabs/centos-presto-test-master'
 INSTALLED_PRESTO_TEST_SLAVE_IMAGE = 'teradatalabs/centos-presto-test-slave'
-DOCKER_MOUNT_POINT = '/mnt/presto-admin'
-LOCAL_MOUNT_POINT = os.path.join(main_dir, 'tmp/docker-pa/')
+DEFAULT_DOCKER_MOUNT_POINT = '/mnt/presto-admin'
+DEFAULT_LOCAL_MOUNT_POINT = os.path.join(main_dir, 'tmp/docker-pa/')
 LOCAL_RESOURCES_DIR = os.path.join(main_dir, 'tests/product/resources/')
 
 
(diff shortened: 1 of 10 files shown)


petroav:
@cawallin & @rschlussel please review

tests/docker_cluster.py
```diff
@@ -44,9 +45,14 @@ class DockerCluster(object):
     for testing purposes.
 
     """
-    def __init__(self, master_host, slave_hosts):
+    def __init__(self, master_host, slave_hosts,
+                 local_mount_dir, docker_mount_dir):
         self.master = master_host
         self.slaves = slave_hosts
+        # the root path for all local mount points; to get a particular
+        # container mount point call get_container_local_mount_dir()
```
rschlussel: I don't see that method.  Is this get_local_mount_dir(self, host)?

tests/product/test_installation.py
```diff
@@ -54,11 +54,12 @@ def test_install_non_root(self):
             sudo -u app-admin tar jxf prestoadmin-*.tar.bz2
             cd prestoadmin
             sudo -u app-admin ./install-prestoadmin.sh
-        """.format(mount_dir=DOCKER_MOUNT_POINT, install_dir=install_dir)
+        """.format(mount_dir=DEFAULT_DOCKER_MOUNT_POINT,
```
rschlussel: this should use self.docker_cluster.default_mount_dir
Same for all the other tests in this file.

tests/product/test_topology.py
```diff
@@ -91,14 +91,14 @@ def test_topology_show_empty_config(self):
...
-        self.copy_to_host(
+        self.docker_cluster.copy_to_host(
             os.path.join(LOCAL_RESOURCES_DIR, 'invalid_json.json'),
             self.docker_cluster.master
         )
         self.docker_cluster.exec_cmd_on_container(
             self.docker_cluster.master,
             "cp %s /etc/opt/prestoadmin/config.json" %
-            os.path.join(DOCKER_MOUNT_POINT, "invalid_json.json")
+            os.path.join(DEFAULT_DOCKER_MOUNT_POINT, "invalid_json.json")
```
rschlussel: this should also use self.docker_cluster.docker_mount_point

tests/docker_cluster.py
petroav: Yep, updated.

tests/product/test_installation.py
petroav: Yes! Great catch!

PR #5, opened 2015-06-30
pa test infra: provide unique hosts to docker containers

petroav:
* Suffix the user given hostnames with a uuid in order to obtain isolation.
  This enables us to concurrently run multiple presto-admin test
  instances.

Task: SWARM-806

### DIFF
--- a/tests/docker_cluster.py
+++ b/tests/docker_cluster.py
@@ -23,6 +23,7 @@ import sys
 import shutil
 import os
 import errno
+import uuid
 from time import sleep
 
 from docker import Client
@@ -47,8 +48,13 @@ class DockerCluster(object):
     """
     def __init__(self, master_host, slave_hosts,
                  local_mount_dir, docker_mount_dir):
-        self.master = master_host
-        self.slaves = slave_hosts
+        # see PyDoc for all_internal_hosts() for an explanation on the
+        # difference between an internal and regular host
+        self.internal_master = master_host
+        self.internal_slaves = slave_hosts
+        self.master = master_host + '-' + str(uuid.uuid4())
+        self.slaves = [slave + '-' + str(uuid.uuid4())
+                       for slave in slave_hosts]
         # the root path for all local mount points; to get a particular
         # container mount point call get_local_mount_dir()
         self.local_mount_dir = local_mount_dir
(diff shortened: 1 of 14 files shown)


petroav:
@cawallin & @rschlussel, please review

tests/docker_cluster.py
```diff
@@ -60,8 +66,34 @@ def __init__(self, master_host, slave_hosts,
...
+
     def get_local_mount_dir(self, host):
-        return os.path.join(self.local_mount_dir, host)
+        return os.path.join(self.local_mount_dir,
+                            self.__get_unique_host(host))
+
+    def __get_unique_host(self, host):
+        matches = [unique_host for unique_host in self.all_hosts()
+                   if unique_host.startswith(host)]
+        if matches:
```
cawallin: So this method is ensuring that get_local_mount_dir will work on either master or master-124ab2312328?

tests/docker_cluster.py
```diff
@@ -319,10 +355,12 @@ def clean_up_presto_test_images(self):
 
     def get_ip_address_dict(self):
         ip_addresses = {}
-        all_hosts = self.all_hosts()
-        for host in all_hosts:
+        for host, internal_host in zip(self.all_hosts(),
+                                       self.all_internal_hosts()):
             inspect = self.client.inspect_container(host)
             ip_addresses[host] = inspect['NetworkSettings']['IPAddress']
+            ip_addresses[internal_host] = \
```
cawallin: Doesn't this overwrite the value?

tests/product/base_product_case.py
```diff
@@ -190,23 +190,25 @@ def build_installer_in_docker(self):
...
+            cluster = self.docker_cluster
         for dist_file in os.listdir(local_dist_dir):
             if fnmatch.fnmatch(dist_file, "prestoadmin-*.tar.bz2"):
-                self.docker_cluster.copy_to_host(
+                cluster.copy_to_host(
                     os.path.join(local_dist_dir, dist_file),
                     dest_host)
 
-    def install_presto_admin(self, host=None):
+    def install_presto_admin(self, cluster, host=None):
```
cawallin: I don't object to this, but why was it necessary/helpful to pass cluster in as an arg?

tests/product/test_installation.py
```diff
@@ -87,7 +87,7 @@ def test_install_from_different_dir(self):
     def test_install_on_wrong_os_offline_installer(self):
         image = 'ubuntu'
         tag = '14.04'
-        host = image + '-' + self.docker_cluster.master
+        host = image + '-master'
```
cawallin: so technically this test couldn't run concurrently?

tests/product/base_product_case.py
```diff
@@ -155,11 +155,11 @@ def build_installer_in_docker(self):
             shutil.copytree(
```
cawallin: trying to comment on line 137... anyway, can you add the uuid to this container too?

rschlussel:
looks good

tests/docker_cluster.py
petroav: yep

tests/docker_cluster.py
petroav: I don't think so. What you end up here is a dict whose keys are the set of all unique hosts + the set of all internal hosts. That way if you query the dict with a unique or internal host, you'll get the right IP. The iteration happens with a tuple: master (being the internal host) and master-alksdjf-afsdgas-asdf (being the unique host).

cawallin: oh, yes. that's true

cawallin:
Looks good

tests/product/base_product_case.py
petroav: This is needed in the case you'd like to install presto-admin on a cluster that is not one of the defaults (just regular centos or centos with presto; when those clusters get provisioned the resulting DockerCluster gets assigned to self.docker_cluster). Examples where this is needed include single node clusters, like the ones needed for the installer tests.

PR #7, opened 2015-07-01
Online offline

petroav:


tests/product/base_product_case.py
```diff
@@ -124,16 +122,25 @@ def setup_docker_cluster(self, cluster_type='centos'):
...
-            self.build_installer_in_docker()
-        return DIST_DIR
+        if 'docker_cluster' in locals():
+            self.docker_cluster.tear_down_containers()
 
-    def build_installer_in_docker(self):
+    def build_dist_if_necessary(self, cluster=None):
+        if not cluster:
+            cluster = self.docker_cluster
+        if (not cluster.get_dist_dir() or
```
rschlussel: If every test has a different dist directory, then we will always need to build dist

tests/product/base_product_case.py
```diff
@@ -124,16 +122,25 @@ def setup_docker_cluster(self, cluster_type='centos'):
...
+                'prestoadmin-*.tar.bz2')
+            ):
+            cluster.clean_up_presto_test_images()
+            self.build_installer_in_docker(cluster=cluster)
+        return cluster.get_dist_dir()
+
+    def build_installer_in_docker(self, online_installer=False, cluster=None):
+        if not cluster:
+            cluster = self.docker_cluster
         container_name = 'installer'
```
rschlussel: If there's another test running that's using the installer container, you won't be able to create another container with the same name.

tests/product/base_product_case.py
```diff
@@ -124,16 +122,25 @@ def setup_docker_cluster(self, cluster_type='centos'):
 
     def tearDown(self):
         self.restore_stdout_stderr_keep_open()
```
rschlussel: We should also delete the dist directory at the end of the test since they aren't reused, so we don't leave all these extra directories lying around.

tests/product/base_product_case.py
petroav: See my previous reply.

tests/product/base_product_case.py
petroav: Since the dist directory is associated with a given DockerCluster object, that directory gets cleaned up in _tear_down_container().

rschlussel: Ah, sorry, didn't see that.

tests/product/base_product_case.py
rschlussel: got it

tests/product/base_product_case.py
petroav: Good point. I'll add a flag to build_dist_if_necessary() and each test will say if it needs a unique installer archive to be created.

PR #9, opened 2015-07-09
Remove temp directories and pin docker-py version

cawallin:
* pin docker-py version to 1.2.2, because docker recently updated the
  client API version which caused tests to fail
* call the tearDown() method of BaseProductTestCase in BaseProductCase
  so that the temp directories are cleaned up

Testing: manually verified temp directories went away


rschlussel:
lgtm

akshatnair:
lgtm

PR #10, opened 2015-07-10
Download RPM earlier to fix Jenkins issue

cawallin:
Some of the earlier tests that don't need to install the Presto RPM
were failing because presto_rpm_filename was not set. We thus download
Presto earlier.

Additionally, change the tox version to 1.9.2 because we also support that.

Testing: make test

### DIFF
--- a/tests/product/base_product_case.py
+++ b/tests/product/base_product_case.py
@@ -91,7 +91,14 @@ task.max-memory=1GB\n"""
             # are multiple RPMs, the last one is probably the latest
             return sorted(rpm_names)[-1]
         else:
-            return None
+            # TODO: once the RPM is on Maven Central, pull the RPM from there
+            rpm_filename = 'presto-0.101-1.0.x86_64.rpm'
+            rpm_path = os.path.join(prestoadmin.main_dir,
+                                    self.presto_rpm_filename)
+            urllib.urlretrieve('http://teradata-download.s3.amazonaws.com/'
+                               'aster/presto/lib/presto-0.101-1.0.x86_64.rpm',
+                               rpm_path)
+            return rpm_filename
 
     def setup_docker_cluster(self, cluster_type='centos'):
         cluster_types = ['presto', 'centos']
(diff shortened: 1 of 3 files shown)


cawallin:
@akshatnair @petroav review please

tests/product/base_product_case.py
```diff
@@ -91,7 +91,14 @@ def detect_presto_rpm(self):
             # are multiple RPMs, the last one is probably the latest
             return sorted(rpm_names)[-1]
         else:
-            return None
+            # TODO: once the RPM is on Maven Central, pull the RPM from there
+            rpm_filename = 'presto-0.101-1.0.x86_64.rpm'
+            rpm_path = os.path.join(prestoadmin.main_dir,
+                                    self.presto_rpm_filename)
+            urllib.urlretrieve('http://teradata-download.s3.amazonaws.com/'
+                               'aster/presto/lib/presto-0.101-1.0.x86_64.rpm',
```
akshatnair: What is this aster s3?

tests/product/base_product_case.py
```diff
@@ -241,17 +248,8 @@ def check_if_corrupted_rpm(self):
...
-            self.presto_rpm_filename = 'presto-0.101-1.0.x86_64.rpm'
-            rpm_path = os.path.join(prestoadmin.main_dir,
-                                    self.presto_rpm_filename)
-            urllib.urlretrieve('http://teradata-download.s3.amazonaws.com/'
-                               'aster/presto/lib/presto-0.101-1.0.x86_64.rpm',
-                               rpm_path)
-        else:
-            rpm_path = os.path.join(prestoadmin.main_dir,
-                                    self.presto_rpm_filename)
+        rpm_path = os.path.join(prestoadmin.main_dir,
```
akshatnair: u already created the rpm_path in the method above, so this might not be needed?

akshatnair:
lgtm mod comment

tests/product/base_product_case.py
cawallin: no, it is necessary; we currently just safe the rpm_filename not the full path (because it's helpful to just have the base path)

PR #11, opened 2015-07-13
Isolate Docker references to DockerCluster

cawallin:
In preparation for adding other types of clusters to presto-admin product
tests, refactor methods that mention "docker" or "containers" to not
do so, and move all uses of docker client into DockerCluster. Add some more
additional cluster arguments to BaseProductTestCase methods, so that
the cluster installation can be done by DockerCluster.

Refactor dump_and_copy_topology to use the write_content_to_host method,
since it was duplicating logic.

Testing: make clean test-all


cawallin:
In the next patch, I'll make a cluster that has the same methods as DockerCluster but which connects to an arbitrary cluster you specify by IP.  @rschlussel @petroav , can you review please?

tests/product/base_product_case.py
```diff
@@ -100,59 +97,51 @@ def detect_presto_rpm(self):
...
+        self.install_presto_admin(cluster=cluster)
+        self.upload_topology(cluster=cluster)
+        self.server_install(cluster=cluster)
+
     def tearDown(self):
         self.restore_stdout_stderr_keep_open()
-        if hasattr(locals()['self'], 'docker_cluster'):
-            self.docker_cluster.tear_down_containers()
+        if self.cluster:
+            self.cluster.tear_down_containers()
```
rschlussel: this still uses containers

cawallin: I'll rename it to tear_down_cluster, which for an arbitrary cluster will uninstall presto and presto-admin.

tests/docker_cluster.py
```diff
@@ -290,7 +290,7 @@ def _are_centos_container_services_up(self, host):
             return False
         return True
 
-    def exec_cmd_on_container(self, host, cmd, raise_error=True, tty=False):
+    def exec_cmd(self, host, cmd, raise_error=True, tty=False):
```
rschlussel: Could you call this exec_cmd_on_host?  Then it will match write_content_to_host.  though the argument order is different.

tests/docker_cluster.py
```diff
@@ -341,12 +357,12 @@ def check_for_presto_images():
 
     def run_script(self, script_contents, host):
```
rschlussel: Then I guess this one would also be run_script_on_host.

tests/docker_cluster.py
cawallin: done

tests/docker_cluster.py
cawallin: done

rschlussel:
looks good

tests/docker_cluster.py
```diff
@@ -300,14 +300,30 @@ def exec_cmd_on_container(self, host, cmd, raise_error=True, tty=False):
...
+            presto_cluster.client.commit(
+                presto_cluster.master,
+                INSTALLED_PRESTO_TEST_MASTER_IMAGE
+            )
+            presto_cluster.client.commit(
+                presto_cluster.slaves[0],
+                INSTALLED_PRESTO_TEST_SLAVE_IMAGE
+            )
+            return presto_cluster
+        else:
```
petroav: Could we ever get to this else? install_func is a required argument so if you call it with no arguments an exception will be raised.

cawallin: but if install_func is null you could get to else. I'd be fine with removing the else, though.

rschlussel: I'm not sure we need to be that defensive here, so I think it's fine to remove it.  but the pythonic way to have this in would be to use a try-catch.

tests/product/base_product_case.py
```diff
@@ -100,59 +97,51 @@ def detect_presto_rpm(self):
...
+    def install_default_presto(self, cluster):
+        self.install_presto_admin(cluster=cluster)
+        self.upload_topology(cluster=cluster)
+        self.server_install(cluster=cluster)
+
     def tearDown(self):
         self.restore_stdout_stderr_keep_open()
-        if hasattr(locals()['self'], 'docker_cluster'):
-            self.docker_cluster.tear_down_containers()
+        if self.cluster:
```
petroav: Can you double check that this works? I feel like I tried something along these lines before going to the more intense solution.

cawallin: I added self.cluster = null at the top, so I think it'll work because it's always defined

PR #12, opened 2015-07-13
Fix tests for the case where the RPM is not yet downloaded

cawallin:
* fixed typo -- needed rpm_filename, not self.presto_rpm_filename which
  may not yet be set

Testing: made sure product tests downloaded RPM correctly


cawallin:
@anusudarsan please review

rschlussel:
lgtm

PR #16, opened 2015-07-21
Add additional startup check for docker

cawallin:
The docker containers sometimes were not fully started, such that when
running the following commands:

cp /mnt/presto-admin/prestoadmin-*.tar.bz2 /home/app-admin
chown app-admin /home/app-admin/prestoadmin-*.tar.bz2

The second command would fail saying "not a directory", because the
/home/app-admin directory had not yet been created, so the tarball
was copied to that location instead.

This commit adds a check that /home/app-admin exists, and improves
debugging messages in other places too.

Testing: tests.product.test_installation


rschlussel:
lgtm

akshatnair:
lgtm

PR #20, opened 2015-07-23
Add connections between all nodes to /etc/hosts of slave docker nodes

cawallin:
* master already had all of the entries for all of the nodes in
  /etc/hosts, but the slaves did not.
* quarantine status tests because they don't work properly now that
  the networking is fixed

Testing: make test-all as sandbox job


cawallin:
@rschlussel 
Should I explain more about why presto didn't work in the commit message or is this sufficient?

PR #22, opened 2015-07-29
Make status rely only on coordinator information

cawallin:
* Query only the coordinator for information in system.runtime.nodes,
  because that information from the workers can be stale and, in the
  worst cases, causes timeouts when you query it.
* unquarantine status tests
* For each of the servers, check if it is running by "service presto
  status" rather than pinging the server
* fix lookup_port to go to a different host properly

Testing: make test-all as sandbox job

### DIFF
--- a/tests/product/test_status.py
+++ b/tests/product/test_status.py
@@ -16,11 +16,11 @@
 Product tests for presto-admin status commands
 """
 import os
-import re
 
 from nose.plugins.attrib import attr
 
-from tests.product.base_product_case import BaseProductTestCase, PRESTO_VERSION
+from tests.product.base_product_case import BaseProductTestCase, \
+    PRESTO_VERSION, PrestoError
 from prestoadmin.util.constants import COORDINATOR_DIR, WORKERS_DIR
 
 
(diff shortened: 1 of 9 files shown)


cawallin:
@anusudarsan @rschlussel can you review?

tests/unit/resources/server_status_out.txt
```diff
@@ -9,9 +9,10 @@ Server Status:
...
 	Node is active: False
+	Connectors:     hive, system, tpch
 Server Status:
 	Node3(IP: IP3 roles: worker): Running
-	No information available
+	No information available: the coordinator has not yet discovered this node
 Server Status:
-	Node4(IP:  roles: worker): Not Running
-	No information available
+	Node4(IP: Unknown roles: worker): Not Running
```
rschlussel: While you're here, could you put a comma before roles in the status output, and maybe capitalize roles too?

tests/product/test_status.py
```diff
@@ -80,9 +80,10 @@ def test_connection_to_coordinator_lost(self):
...
             self.cluster.slaves[0])
-        status_output = self.run_prestoadmin('server status')
+        status_output = self._server_status_with_retries()
         statuses = self.node_not_available_status(
-            ips, topology, self.cluster.internal_slaves[0])
+            ips, topology, self.cluster.internal_slaves[0],
+            coordinator_down=True)
         self.check_status(status_output, statuses)
 
     @attr('quarantine')
```
cawallin: need to actually unquarantine these

rschlussel:
otherwise looks good

tests/unit/resources/server_status_out.txt
cawallin: sure, will do

PR #25, opened 2015-07-31
pa-tests: Switch to pull RPM artifact from Maven

cawallin:
Testing: make test-all on sandbox

### DIFF
--- a/tests/product/base_product_case.py
+++ b/tests/product/base_product_case.py
@@ -32,7 +32,9 @@ from tests.configurable_cluster import ConfigurableCluster
 from tests.docker_cluster import DockerCluster, DockerClusterException, \
     LOCAL_RESOURCES_DIR, DEFAULT_LOCAL_MOUNT_POINT, DEFAULT_DOCKER_MOUNT_POINT
 
-PRESTO_RPM_GLOB = r'presto-*.x86_64.rpm'
+RPM_BASENAME = r'presto.*'
+
+PRESTO_RPM_GLOB = r'presto*.rpm'
 PRESTO_VERSION = r'presto-main:.*'
 RETRY_TIMEOUT = 120
 RETRY_INTERVAL = 5
@@ -98,13 +100,13 @@ task.max-memory=1GB\n"""
             # are multiple RPMs, the last one is probably the latest
             return sorted(rpm_names)[-1]
         else:
-            # TODO: once the RPM is on Maven Central, pull the RPM from there
-            rpm_filename = 'presto-0.101-1.0.x86_64.rpm'
+            rpm_filename = 'presto-server-rpm.rpm'
             rpm_path = os.path.join(prestoadmin.main_dir,
                                     rpm_filename)
-            urllib.urlretrieve('http://teradata-download.s3.amazonaws.com/'
-                               'aster/presto/lib/presto-0.101-1.0.x86_64.rpm',
-                               rpm_path)
+            urllib.urlretrieve(
+                'https://repository.sonatype.org/service/local/artifact/maven'
+                '/content?r=central-proxy&g=com.facebook.presto'
+                '&a=presto-server-rpm&e=rpm&v=RELEASE', rpm_path)
             return rpm_filename
 
     def setup_cluster(self, cluster_type='base'):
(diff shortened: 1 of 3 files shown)


akshatnair:
lgtm

cawallin:
@akshatnair @rschlussel  -- I updated this some, can you review it?

tests/product/base_product_case.py
```diff
@@ -98,13 +98,13 @@ def detect_presto_rpm(self):
             # are multiple RPMs, the last one is probably the latest
             return sorted(rpm_names)[-1]
         else:
-            # TODO: once the RPM is on Maven Central, pull the RPM from there
-            rpm_filename = 'presto-0.101-1.0.x86_64.rpm'
+            rpm_filename = 'presto-server-rpm.rpm'
```
rschlussel: so the rpm name will never have anything else in it?  (x86_64, version number, etc.)

tests/product/base_product_case.py
```diff
@@ -335,16 +335,23 @@ def assert_installed(self, container, cluster=None, msg=None):
             cluster = self.cluster
         try:
             check_rpm = cluster.exec_cmd_on_host(
-                container, 'rpm -q presto')
-            self.assertEqual(self.presto_rpm_filename[:-4] + '\n', check_rpm,
-                             msg=msg)
+                container, 'rpm -q presto-server-rpm')
+            self.assertRegexpMatches(
+                check_rpm, self.presto_rpm_filename[:-4] + r'.*\n', msg=msg
```
rschlussel: Is there a way to make it more clear what this is doing?

tests/product/base_product_case.py
```diff
@@ -417,7 +424,7 @@ def replace_keywords(self, text, cluster=None, **kwargs):
             cluster = self.cluster
         test_keywords = {
             'rpm': self.presto_rpm_filename,
-            'rpm_basename': self.presto_rpm_filename[:-4],
+            'rpm_basename': self.presto_rpm_filename[:-4] + '.*',
```
rschlussel: what's the rpm_filename/rpm_basename supposed to represent?  I'm confused why you would add a .* after you get rid of the .rpm from the rpm_filename.

tests/product/base_product_case.py
```diff
@@ -417,7 +424,7 @@ def replace_keywords(self, text, cluster=None, **kwargs):
...
             'rpm_basename_without_arch': self.presto_rpm_filename[:-11],
```
rschlussel: Is this always going to be -11?  Is it because we always use the rpm for the same arch?

tests/product/base_product_case.py
cawallin: it wasn't clear to me how to get this url library to save the original rpm file name from the download

tests/product/base_product_case.py
cawallin: there are extra things (e.g. -0:0.113.SNAPSHOT-1.x86_64 or something) which might change which we don't care about

tests/product/base_product_case.py
cawallin: we always build x86_64, yes.

tests/product/base_product_case.py
cawallin: see above

tests/product/base_product_case.py
rschlussel: I understand the goal, I just think the construction of the regex is a bit contrived.  The rpm name isn't actually tied to the filename at all.  It happens to relate, but you made up the name of the file when you downloaded it.  Maybe put 'presto-server-rpm.*' as its own constant or build them both from a constant that's just 'presto-server-rpm'?

tests/product/base_product_case.py
cawallin: sure, I can do that

tests/product/base_product_case.py
rschlussel: got it

tests/product/base_product_case.py
rschlussel: excellent, thanks!

rschlussel:
The rest looks good

PR #26, opened 2015-08-03
Update Docker to pull one version of centos-ssh

cawallin:
Testing: Deleted all of the images and made sure that only
centos-6-1.2.0 was pulled and that a few tests ran.


cawallin:
@jreslock @petroav -- review please, this might help with the automation (or if it doesn't, it will save a TON of space since it won't download 6 different versions of centos-ssh)

petroav:
looks good

PR #27, opened 2015-08-03
Update default Presto configuration

cawallin:
* add -DHADOOP_USER_NAME=hive in jvm.config
* add node-scheduler.include-coordinator = false, because the default
  is to do work on the coordinator
* Check if discovery-uri starts with http://

Testing: make clean lint test test-all

### DIFF
--- a/prestoadmin/coordinator.py
+++ b/prestoadmin/coordinator.py
@@ -28,23 +28,28 @@ from prestoadmin.presto_conf import validate_presto_conf, get_presto_conf
 from prestoadmin.util import constants
 from prestoadmin.util.exception import ConfigurationError
 
-DEFAULT_PROPERTIES = {"node.properties":
-                      {"node.environment": "presto",
-                       "node.data-dir": "/var/lib/presto/data",
-                       "plugin.config-dir": "/etc/presto/catalog",
-                       "plugin.dir": "/usr/lib/presto/lib/plugin"},
-                      "jvm.config": ["-server",
-                                     "-Xmx1G",
-                                     "-XX:-UseBiasedLocking",
-                                     "-XX:+UseG1GC",
-                                     "-XX:+ExplicitGCInvokesConcurrent",
-                                     "-XX:+HeapDumpOnOutOfMemoryError",
-                                     "-XX:+UseGCOverheadLimit",
-                                     "-XX:OnOutOfMemoryError=kill -9 %p"],
-                      "config.properties": {"coordinator": "true",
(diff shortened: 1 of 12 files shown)


cawallin:
@rschlussel can you take a look please?

prestoadmin/coordinator.py
```diff
@@ -82,7 +85,14 @@ def build_defaults():
 
 def validate(conf):
     validate_presto_conf(conf)
-    if conf["config.properties"]["coordinator"] != "true":
-        raise ConfigurationError("Coordinator cannot be false in the "
-                                 "coordinator's config.properties.")
+    if conf['config.properties']['coordinator'] != 'true':
+        raise ConfigurationError('Coordinator cannot be false in the '
+                                 'coordinator\'s config.properties.')
+    if 'discovery.uri' not in conf['config.properties']:
```
rschlussel: could you add negative tests for discovery.uri missing/not http?

cawallin: I think I added one negative test for not http, but I'll add one for missing

prestoadmin/workers.py
```diff
@@ -71,27 +72,30 @@ def _get_conf():
...
-                "is not.  The default discovery-uri is "
-                "http://<coordinator>:8080")
+        raise ConfigurationError(
+            'discovery.uri should not be localhost in a '
+            'multi-node cluster, but found ' + urlparse.urlunparse(uri) +
+            '.  You may have encountered this error by '
+            'choosing a coordinator that is localhost and a worker that '
+            'is not.  The default discovery-uri is '
+            'http://<coordinator>:8080')
+    elif uri.scheme != 'http':
```
rschlussel: would it make sense to move the check for http start to validate_presto_conf since it is common to both coordinator and workers.  Also, this one should also check that discovery.uri exists.

prestoadmin/coordinator.py
cawallin: actually, tested both for coordinator but neither for worker. fixed it!

prestoadmin/workers.py
cawallin: didn't quite know that existed! moved the check there

cawallin:
Updated

rschlussel:
lgtm

cawallin:
@rschlussel: I added query.max-memory and query.max-memory-per-node, mind taking a quick look?

rschlussel:
lgtm

cawallin:
merged

PR #29, opened 2015-08-05
Update tests based on new configs

cawallin:
A few things broke with the new configs added in a previous commit.
1. Add the ability to add extra configs in addition to the test configs
2. annotate write_test_configs() with @nottest so it isn't treated as a test
3. With the config node-scheduler.include-coordinator=false, we can't
   return node information if only the coordinator is running. Update test
   and status behavior.
4. Add a retry for test_collect_query_info, because it now needs to wait
   until the coordinator discovers all of the workers.

Testing: make test-all as a sandbox build

### DIFF
--- a/tests/product/test_status.py
+++ b/tests/product/test_status.py
@@ -15,13 +15,11 @@
 """
 Product tests for presto-admin status commands
 """
-import os
 
 from nose.plugins.attrib import attr
 
 from tests.product.base_product_case import BaseProductTestCase, \
     PRESTO_VERSION, PrestoError
-from prestoadmin.util.constants import COORDINATOR_DIR, WORKERS_DIR
 
 
 class TestStatus(BaseProductTestCase):
@@ -48,7 +46,9 @@ class TestStatus(BaseProductTestCase):
 
         # Test with worker not started
         self.run_prestoadmin('server start -H master')
-        status_output = self._server_status_with_retries()
+        # don't run with retries because it won't be able to query the
+        # coordinator because the coordinator is set to not be a worker
+        status_output = self.run_prestoadmin('server status')
         self.check_status(
             status_output,
             self.single_node_up_status(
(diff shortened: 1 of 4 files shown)


cawallin:
@rschlussel can you review this please? Also, let me know if the commit message is incoherent.

cawallin:
@gazing-star -- I don't expect you to have any context on this, but you might as well look at this PR to see our process of code review and then merging the commit into the main codebase :)

tests/product/test_status.py
```diff
@@ -48,11 +46,13 @@ def test_status_happy_path(self):
...
         self.run_prestoadmin('server start -H master')
-        status_output = self._server_status_with_retries()
+        # don't run with retries because it won't be able to query the
+        # coordinator because the coordinator is set to not be a worker
+        status_output = self.run_prestoadmin('server status')
         self.check_status(
             status_output,
             self.single_node_up_status(
-                ips, self.cluster.internal_master))
+                ips, self.cluster.internal_master, coordinator_down=True))
```
rschlussel: could you change the name of coordinator_down since the coordinator isn't down?

cawallin: sure. it seems like we always want to have this be true anyway, so I can clean it up and change that.

tests/product/test_status.py
```diff
@@ -48,11 +46,13 @@ def test_status_happy_path(self):
 
         # Test with worker not started
```
rschlussel: In a next patch could you move the tests that aren't actually happy path tests into their own methods?  It would make it easier to quickly diagnose jenkins issues.

tests/product/test_status.py
rschlussel: excellent, thanks!

rschlussel:
In the commit message could you just make clear that the items in your list aren't the things that broke, but rather the fixes?

tests/product/test_status.py
cawallin: Will do

cawallin:
@rschlussel : updated. I'll add another commit that separates the happy path tests.

rschlussel:
lgtm

PR #34, opened 2015-08-11
Add debugging for Docker cluster issue

cawallin:
In order to figure out an intermittent RPM db error, take a snapshot of
the cluster when an error occurs so that the issues can be investigated.

### DIFF
--- a/tests/product/base_product_case.py
+++ b/tests/product/base_product_case.py
@@ -398,6 +398,11 @@ task.max-memory=1GB\n"""
                 check_rpm, RPM_BASENAME + '\n', msg=msg
             )
         except OSError as e:
+            if isinstance(cluster, DockerCluster):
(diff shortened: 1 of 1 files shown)


tests/product/base_product_case.py
```diff
@@ -398,6 +398,11 @@ def assert_installed(self, container, cluster=None, msg=None):
                 check_rpm, RPM_BASENAME + '\n', msg=msg
             )
         except OSError as e:
+            if isinstance(cluster, DockerCluster):
```
rschlussel: Is this just a temporary change?  I'm worried about cluttering up the system running the tests and forgetting about these images lying around

cawallin: yes. sorry for not mentioning that.

rschlussel:
k looks good

PR #38, opened 2015-08-20
presto-admin: make presto-admin parse teradata versions

arhimondr:
Now we are checking that presto is installed before checking version using 'rpm -q presto' and 'rpm -q presto-server-rpm'.
So, now we will get 'Presto is not installed' message just in case when presto is not installed at all.

Then if presto is installed, we are trying to parse presto version usin '(\d+)\.(\d+)t?([-\.]SNAPSHOT)?' regexp.
First group - is the major version, second group is the minor version.
This regexp should parse all possible versions including teradata ones and snapshot ones.

When version doesn't match this pattern we are showing to user 'Incorrect presto version' message
instead of telling him that presto is not installed at all.

Testing: manual

### DIFF
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -51,7 +51,7 @@ NODE_INFO_PER_URI_SQL = 'select http_uri, node_version, active from ' \
 EXTERNAL_IP_SQL = 'select url_extract_host(http_uri) from system.runtime.nodes' \
                   ' WHERE node_id = \'%s\''
 CONNECTOR_INFO_SQL = 'select catalog_name from system.metadata.catalogs'
-PRESTO_RPM_MIN_REQUIRED_VERSION = 100
+PRESTO_RPM_MIN_REQUIRED_VERSION = 103
 PRESTO_TD_RPM = ['101t']
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


akshatnair:
Can you please run "make test" to check unit tests are passing and also run product tests for install and uninstall?

cawallin:
@ilfrin: presto-admin is meant to work with all version of Presto -- 115, 115t, and any SNAPSHOT version.

@arhimondr: Please also run make lint, if you haven't already, and the status product tests.

ilfrin:
@cawallin I know, but my point is now if you say it needs 115 if will accept ANY 115 flavor, there is no way of stating that I want a vanilla 115 or a TD 115 build.

prestoadmin/server.py
```diff
@@ -219,34 +219,51 @@ def check_presto_version():
...
+    matched = re.search('(\d+)\.(\d+)t?([-\.]SNAPSHOT)?', version)
+    if not matched:
+        incorrect_version_str = 'Incorrect presto version:  %s,' % version
+        warn(incorrect_version_str)
+        return incorrect_version_str
+
+    major_version = matched.group(1)
+    minor_version = matched.group(2)
+
+    assert major_version.isdigit(), 'incorrect major presto version'
```
cawallin: isn't the major version and minor version guaranteed to be a digit because of the regex you use? If it's necessary to check this, it should be with a warning like above and not an assertion

prestoadmin/server.py
```diff
@@ -219,34 +219,51 @@ def check_presto_version():
...
+    assert minor_version.isdigit(), 'incorrect minor presto version'
+
+    if int(minor_version) < PRESTO_RPM_MIN_REQUIRED_VERSION:
```
cawallin: I know it's a little tangential, but can you update PRESTO_RPM_MIN_REQUIRED_VERSION for this presto-admin to 103 instead of 100? That will make it clear that you need at least presto version 103 for this version of presto-admin because of config variables that were added in that Presto release.

cawallin:
@ilfrin All this check is doing is seeing if the Presto version is compatible with this version of presto-admin. Since there aren't breaking changes between 0.115 and 0.115t, this is fine.  Hopefully there will never be changes that are that significant, but if there are we can do version checking differently then.

cawallin:
looks good % changes

ilfrin:
@cawallin got it, thanks, we'll revisit this if it creates any issues.

prestoadmin/server.py
arhimondr: done

prestoadmin/server.py
arhimondr: Yes, it is guaranteed by regexp, but when someone accidentally changed the REGEXP it would be better to have immediate fail instead of just warning.

cawallin: warnings in Fabric are the equivalent to immediately failing, but display better to the user in the (hopefully non-existent) chance that we don't catch this before a release

arhimondr: actually it should never happens, because of the reg-exp. It is just some kind of paranoia-check :-) if you don't like it i would better remove it at all =)

cawallin: yeah, why don't you just remove it. thanks

arhimondr:
@cawallin 
Lint, unit tests and product tests are passing. I'm going to merge it into the master.

ilfrin:
Thank you Andrii.

czw., 20 sie 2015, 22:39 Andriy Rosa użytkownik <notifications@github.com>
napisał:

PR #39, opened 2015-08-20
Change init command to /etc/init.d/presto

rschlussel:
Change init command from /etc/rc.d/init.d/presto to /etc/init.d/presto

Testing: make test.  test_control, test_server_uninstall

### DIFF
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -41,7 +41,7 @@ import util.filesystem
 
 __all__ = ['install', 'uninstall', 'start', 'stop', 'restart', 'status']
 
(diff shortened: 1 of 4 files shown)


cawallin:
Looks good -- but does the latest release of Presto from maven have your change? I guess since it's a symlink it should work regardless?

rschlussel:
Latest release does not have my change.  It's been merged but a new release hasn't come out yet.  Since the directory in centos a symlink it should work regardless.  And when you run tests on the appliance, it won't be a symlink, so you'll need this (and an rpm that has my rpm change)

PR #40, opened 2015-08-24
Assorted fixes for automation failures

cawallin:
* Increase the timeout because sometimes there are timeouts because
  the presto-admin commands take longer than expected
* retry RPM download once, since sometimes Maven resets the connection while
  the download is taking place
* Add another yum command to the Dockerfile to try to fix RPM db errors
  (see http://serverfault.com/questions/637549/epel-repo-for-centos-6-causing-error).
  It's unclear if that command will help, but it could.

Testing: sandbox run of a few tests

### DIFF
--- a/tests/docker_cluster.py
+++ b/tests/docker_cluster.py
@@ -60,7 +60,7 @@ class DockerCluster(object):
         # container mount point call get_local_mount_dir()
         self.local_mount_dir = local_mount_dir
         self.mount_dir = docker_mount_dir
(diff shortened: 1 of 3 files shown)


cawallin:
@rschlussel @jreslock @petroav -- if you have any other suggestions let me know; I think that at least the RPM retry should be helpful since 6-7 of the failures last week were caused by RPM download errors

rschlussel:
lgtm

PR #43, opened 2015-08-25
Don't modify /etc/hosts to simulate a node being down

cawallin:
* Change topology to have an invalid IP address for a node that is
  down, rather than modifying /etc/hosts.
* Had to modify some test results

Testing: make clean test-all on both cluster and docker

### DIFF
--- a/tests/docker_cluster.py
+++ b/tests/docker_cluster.py
@@ -169,6 +169,9 @@ class DockerCluster(object):
         self.client.stop(container_name)
         self.client.wait(container_name)
 
+    def get_down_hostname(self, host_name):
+        return host_name
+
     def _remove_host_mount_dirs(self):
         for container_name in self.all_hosts():
             try:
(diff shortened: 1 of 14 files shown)


tests/docker_cluster.py
```diff
@@ -67,7 +67,7 @@ def __init__(self, master_host, slave_hosts,
     def all_hosts(self):
         return self.slaves + [self.master]
 
-    def all_internal_hosts(self):
+    def all_internal_hosts(self, stopped_host=None):
```
rschlussel: is stopped_host used?

cawallin: it's used in configurable_cluster; wanted the method signature to be the same

cawallin: It's not necessary. I removed it.

cawallin: oh, I remember why I had it. it's because you don't know which host is the stopped host and the name is wrong.

PR #46, opened 2015-08-26
Add one-step presto upgrade to presto-admin

ebd2:
This pull request adds one-step upgrade for prestodb to presto-admin. In order to support this, it adds functionality to gather the existing presto configuration from the hosts in the cluster to a local directory, and to deploy configuration to the cluster from a directory. This ensures that the config on the cluster remains the same across an upgrade in the following situations:
* A user has changed the configuration on the cluster outside of presto-admin
* prestodb was installed and configured manually, and presto-admin has no config for the cluster.

The upgrade process works as follows:
1. Gather the cluster config to a local directory.
2. Uninstall presto from the cluster.
3. Install the new presto package on the cluster.
4. Deploy the previous configuration from the local directory.


ebd2:
@cawallin, @rschlussel please review.

prestoadmin/server.py
```diff
@@ -125,6 +127,15 @@ def uninstall():
         print('Package uninstalled successfully on: ' + env.host)
 
 
+@task
+@requires_topology
+def upgrade(local_package_path, local_config_dir):
+    configure_cmds.gather_directory(local_config_dir)
```
rschlussel: instead of gathering the directory content to the presto-admin node and later redeploying it, it would be more efficient to copy the configuration to a temporary location on each node.

prestoadmin/server.py
```diff
@@ -125,6 +127,15 @@ def uninstall():
...
+    uninstall()
```
rschlussel: could you change this to do a package upgrade (i.e. ultimately call out to rpm -u).  This would allow us to be smarter about the things we erase or don't erase depending on if it's an upgrade.  Right now we don't try to do anything smart, but it would probably be good to do in the future.

prestoadmin/configure_cmds.py
```diff
@@ -69,15 +75,59 @@ def deploy(rolename=None):
             abort("Invalid Argument. Possible values: coordinator, workers")
 
 
-def configuration_show(file_name, should_warn=True):
-    file_path = os.path.join(constants.REMOTE_CONF_DIR, file_name)
-    if not files.exists(file_path):
+@task
+@requires_topology
```
cawallin: I don't think this should be a task -- tasks are user-facing, and the above deploy() method is very similar to this, except for specifying the directory.

prestoadmin/server.py
```diff
@@ -125,6 +127,15 @@ def uninstall():
         print('Package uninstalled successfully on: ' + env.host)
 
 
+@task
+@requires_topology
+def upgrade(local_package_path, local_config_dir):
```
cawallin: Tasks all need to have pydoc like in configuration deploy, because it's user facing (when you do ./presto-admin -d configuration deploy, for example, it prints out the pydoc)

prestoadmin/configure_cmds.py
```diff
@@ -33,9 +37,11 @@
 JVM_CONFIG = "jvm.config"
 NODE_PROPERTIES = "node.properties"
 
+ALL_CONFIG = [CONFIG_PROPERTIES, LOG_PROPERTIES, JVM_CONFIG, NODE_PROPERTIES]
+
 _LOGGER = logging.getLogger(__name__)
 
-__all__ = ['deploy', 'show']
+__all__ = ['deploy', 'show', 'gather_directory', 'deploy_directory']
```
rschlussel: I'm not sure gather_directory and deploy_directory should be user facing tasks.

rschlussel:
Could you add product tests too?

prestoadmin/server.py
ebd2: Per discussion, we're going to leave this because it will allow us to handle required configuration changes on upgrade in the future.

prestoadmin/server.py
ebd2: I've added the pydoc, please have a look and make sure you agree with the wording.

ebd2:
I'll be out of the office for several weeks. The work that remains to be done before pushing is as follows:
1. Product tests. Per discussion with @cawallin and @rschlussel, we'd like to add a test that upgrades to a mostly-empty dummy rpm of presto-server. This is to avoid pulling another ~400 MB .rpm over the network just for the sake of a test. The dummy rpm and subsequent verification should ensure that the following scenarios are tested:
  * a file is added
  * a file is removed
  * a file is modified
I found the following guide to creating rpm packages, but didn't make any headway on creating the dummy package: https://fedoraproject.org/wiki/How_to_create_an_RPM_package. A forum post I read advised against creating .rpms on Ubuntu because of possible incompatibilities. I haven't gotten far enough to confirm this, but rpm -qpi presto-server-rpm... returns an error on Ubuntu, which suggests that the warning has merit.
2. The change to use rpm -U is currently broken because rpm -U installs the new package, then uninstalls the previous package. This would work, but the postuninstall script for the package unconditionally removes /etc/presto and /var/lib/presto. There are comments in the script explaining the rationale for the behavior. /etc/presto can be recreated before we push the saved config back out to the cluster, but we should probably look into the behavior of the postuninstall script and see if it should be updated to do something different.

cawallin:
@rschlussel -- can you review please?

cawallin:
noteworthy changes:
* there is now a default temporary config storage place
* connectors are saved as well
* The fabric_api by rolename method was no longer necessary, so I removed it

tests/product/test_server_upgrade.py
```diff
@@ -0,0 +1,120 @@
...
+        self.assert_started(process_per_host)
+
+    def assert_upgraded(self, hosts):
+        for container in hosts:
+            # Still should have the same configs
+            self.assert_installed(container)
+            self.assert_has_default_config(container)
+            self.assert_has_default_connector(container)
+
+            # However, the new RPM removes /usr/lib/presto/lib and
```
rschlussel: could you make clear that these expectations are specifically about the test RPM (not that upgrading in general deletes the plugin directory)

tests/product/test_server_upgrade.py
```diff
@@ -0,0 +1,120 @@
...
+            # /usr/shared/doc/presto/README.txt
+            self.assert_file_content_regex(
+                container,
+                '/usr/shared/doc/presto/README.txt',
+                r'.*New line of text here.$'
+            )
+
+    def copy_upgrade_rpm_to_cluster(self):
+        self.cluster.copy_to_host(
+            os.path.join(LOCAL_RESOURCES_DIR, 'dummy-rpm.rpm'),
```
rschlussel: you need to commit the dummy-rpm.

tests/product/test_server_upgrade.py
```diff
@@ -0,0 +1,120 @@
...
+        self.copy_presto_rpm_to_master()
+        topology = {"coordinator": self.cluster.internal_slaves[0],
+                    "workers": [self.cluster.internal_master,
+                                self.cluster.internal_slaves[1],
+                                self.cluster.internal_slaves[2]]}
+        self.upload_topology(topology=topology)
+        self.cluster.stop_host(
+            self.cluster.slaves[0])
+
+        actual_out = self.server_install()
```
rschlussel: should this be upgrade?

tests/product/test_server_upgrade.py
```diff
@@ -0,0 +1,120 @@
...
+        self.assertRegexpMatches(
+            actual_out,
+            self.down_node_connection_error(self.cluster.internal_slaves[0])
+        )
+
+        for container in [self.cluster.master,
```
rschlussel: don't you need to check everything from assert_upgraded?

tests/product/test_server_upgrade.py
cawallin: sure

tests/product/test_server_upgrade.py
cawallin: thanks, will do

tests/product/test_server_upgrade.py
cawallin: oh. this test is irrelevant. the above test tests the same thing

rschlussel:
lgtm

PR #47, opened 2015-09-01
Update defaults so that Presto starts

cawallin:
With query.max-memory-per-node, Presto preemptively performs a memory check.
Since the max JVM memory was 1GB, there wasn't 1GB free for
query.max-memory-per-node.

Testing: make test, changed product test results


cawallin:
@anusudarsan @mattsfuller can you review please?

anusudarsan:
looks good. what is the err you see ? the launcher.log would show some errors?

cawallin:
The Presto server doesn't start without this patch; the error message is in server.log

PR #54, opened 2015-09-16
Fix build on MacOSX

losipiuk:


tests/unit/test_bdist_prestoadmin.py
```diff
@@ -58,8 +59,9 @@ def test_initialize(self):
         self.assertEquals(bdist.online_install, False)
 
     def test_finalize(self):
-        self.assertEquals(self.bdist.bdist_dir,
-                          'build/bdist.linux-x86_64/prestoadmin')
+        self.assertRegexpMatches(
+            self.bdist.bdist_dir,
+            'build/bdist\.(linux.*|macos.*)/prestoadmin')
```
cawallin: maybe we shouldn't check the operating system part of the filename at all, since this would fail if you ran the tests on windows

cawallin:
one small comment, but looks good

losipiuk:
AC
Merging

PR #55, opened 2015-09-16
Fix product tests to support boot2docker

losipiuk:


cawallin:
as long as this doesn't break the non boot-to-docker workflow, looks good (let me know if you want me to pull the patch and test it out)

losipiuk:
Lets merge it and see automation. I do not expect any problems with that.

cawallin:
ok, that's fine

PR #56, opened 2015-09-16
Create new docker client for each call

losipiuk:
When communicating with docker over http (this is the case when
boot2docker is in use) command could fail if multiple streaming
commands were executed using same docker client.
The request then failed with "Hijack is incompatible with use of
CloseNotifier" error.
Workaround for that (until docker-py library is fixed) is to create
new docker client everytime we need one.

### DIFF
--- a/tests/docker_cluster.py
+++ b/tests/docker_cluster.py
@@ -61,14 +61,20 @@ class DockerCluster(object):
         # container mount point call get_local_mount_dir()
         self.local_mount_dir = local_mount_dir
         self.mount_dir = docker_mount_dir
-
-        kwargs = kwargs_from_env()
-        kwargs['tls'].assert_hostname = False
-        kwargs['timeout'] = 240
-        self.client = Client(**kwargs)
+        self._client = None
 
         self._DOCKER_START_TIMEOUT = 30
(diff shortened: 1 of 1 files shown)


tests/docker_cluster.py
```diff
@@ -61,14 +61,20 @@ def __init__(self, master_host, slave_hosts,
...
         self.mount_dir = docker_mount_dir
+        self._client = None
 
+        self._DOCKER_START_TIMEOUT = 30
+        DockerCluster.__check_if_docker_exists()
+
+    @property
+    def client(self):
+        if self._client:
+            self._client.close()
```
cawallin: if the issue is multiple clients streaming, doesn't closing the client here cause problems?

cawallin: or is it multiple subsequent streaming http commands?

losipiuk: problem appears when one client instance is used for multiple http requests with streaming. So we create new client each time `self.client` is accessed. It could still fail if somebody stores client on local variable. But I do not think that is the case anywhere currently.

cawallin: no, self.client isn't ever stored

cawallin:
looks good

PR #57, opened 2015-09-16
Update version of docker-py to 1.4.0

losipiuk:
This is nothing crucial.
Just result of searching around trying to make tests run on boot2docker.
Does not seem to break anything though. You may merge it some time later if you like.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,12 +2,12 @@ flake8==2.4.0
 mock==1.0.1
 py==1.4.26
 Sphinx>=1.3.1
-tox==1.9.2
+tox==2.1.1
 virtualenv==12.0.7
-wheel==0.23.0
+wheel==0.24.0
 fabric==1.10.1
 requests==2.7.0
-docker-py==1.2.2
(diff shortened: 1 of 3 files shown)


rschlussel:
I think we were intentionally on an older version because the newer version was causing some failure (I don't remember exactly what).  Could you run all the tests with this?

cawallin:
yeah, here's the PR: https://github.com/prestodb/presto-admin/pull/9.  It has to do with both the version of docker you have installed and the client version, so if this doesn't help the boot2docker stuff it'd be better for this change not to go in

losipiuk:
I am currently on newest docker. Updating docker-py was just a try - it did not help - so I will not be forcing update. Probably after the release we can think of freshening libraries up. Users will probably have newest docker installed and at some point old docker-py may stop working with that.

ebd2:
I think we can close this one; I updated to docker-py 1.5.0 a little bit ago :-)

cawallin:
sure, go ahead

PR #59, opened 2015-09-17
Get wheel name for script after running bdist_wheel

cawallin:
For versions of wheel >= 0.25, the incorrect name was being returned from
build_wheel because the archive name is finalized only after running
the command. The cmd.finalize_options() command is not strictly speaking
necessary, but helps with unit tests and ensures that if in the future, cmd
does not get updated when the 'bdist_wheel' command is run, the wheel
name will still be correct.

Testing: make test with wheel version 0.23 and 0.25


cawallin:
@rschlussel @mattsfuller 
by the way, the thing that was causing product tests to fail is that when we create the dist, we call "pip install --upgrade wheel", and the latest version of wheel does not work.

rschlussel:
lgtm

PR #63, opened 2015-09-22
Add test that presto is started as presto user

cawallin:
* Make sure that the user of server's pid is presto
* change run_prestoadmin_script to run_script_from_prestoadmin_dir and
  extend to be able to run on other hosts than master

Testing: make lint, ran test, sandbox with all tests and new RPM

### DIFF
--- a/tests/product/base_product_case.py
+++ b/tests/product/base_product_case.py
@@ -348,17 +348,21 @@ query.max-memory=50GB\n"""
             raise_error=raise_error
         )
 
-    def run_prestoadmin_script(self, script_contents, **kwargs):
+    def run_script_from_prestoadmin_dir(self, script_contents, host='',
+                                        **kwargs):
+        if not host:
+            host = self.cluster.master
+
         script_contents = self.replace_keywords(script_contents,
                                                 **kwargs)
         temp_script = '/opt/prestoadmin/tmp.sh'
         self.cluster.write_content_to_host(
             '#!/bin/bash\ncd /opt/prestoadmin\n%s' % script_contents,
-            temp_script, self.cluster.master)
+            temp_script, host)
         self.cluster.exec_cmd_on_host(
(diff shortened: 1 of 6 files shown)


cawallin:
This patch will need to be added to a special release branch, because it only works with 115t (and the fix in Presto will take a bit to be merged).

cawallin:
@rschlussel @anusudarsan please review

tests/product/test_control.py
```diff
@@ -268,6 +268,19 @@ def test_start_restart_config_file_error(self):
         self.assert_simple_server_restart(expected_restart,
                                           expected_stop=expected_stop)
 
+    def test_started_with_presto_user(self):
+        # note, will only work with 0.115t RPM
+        self.setup_cluster('presto')
```
anusudarsan: can we extract 'presto' to a const?

anusudarsan:
looks good

tests/product/base_product_case.py
```diff
@@ -348,17 +348,21 @@ def run_prestoadmin(self, command, raise_error=True, cluster=None):
             raise_error=raise_error
         )
 
-    def run_prestoadmin_script(self, script_contents, **kwargs):
+    def run_script_from_prestoadmin_dir(self, script_contents, host='',
```
rschlussel: why don't you give host a default value of self.cluster.master?

rschlussel: oh can it not get self as a value?

rschlussel:
looks good

tests/product/base_product_case.py
cawallin: yeah, it can't have self in the default

tests/product/test_control.py
cawallin: good idea, I'll do this in a different commit though

PR #64, opened 2015-09-23
Call splitlines before passing output to get_process_per_host.

ebd2:
Ensure splitlines is called before passing command output to get_process_per_host. This ensures that we are correctly checking if the server is started.

### DIFF
--- a/tests/product/base_product_case.py
+++ b/tests/product/base_product_case.py
@@ -469,6 +469,13 @@ query.max-memory=50GB\n"""
 
     def get_process_per_host(self, output_lines):
         process_per_host = []
+        # We found some places where we were incorrectly passing a string
+        # containing the output rather than an iterable collection of lines.
+        # Since strings don't have an __iter__ attribute, we can catch this
+        # error.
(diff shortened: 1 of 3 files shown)


cawallin:
oh, it was only a few places? excellent. can you change the name of get_process_per_host to suggest that you have to have splitlines?
looks good, though make sure to run ``make lint``

ebd2:
Yeah, there were only 16 uses of get_process_per_host in the product tests, and most of them were already correct. ```make lint``` comes up clean.

Per offline conversation, I'm going to make get_process_per_host determine if it has been passed a string by checking hasattr(output_lines, '__iter__') and raising an Error if that returns false. Strings are iterable, just like lists, but they don't have an '__iter__' attribute, so we can distinguish them from lists that way. We're in agreement that this isn't terribly pythonic, but it does make it harder to reintroduce bugs.

cawallin:
looks good

cawallin:
@ebd2 you pushed this right? if so, you should close the PR

ebd2:
@cawallin I did. Sorry about that.

PR #67, opened 2015-10-06
Archive generated by bdist_prestoadmin should not include .

ebd2:
This change adds a base_dir argument to make_archive so that the archives we
generate do not include an entry for the current directory (.). Without this
change, we apply the permissions of some arbitrary directory on the machine
where the archive is built to the parent directory of prestoadmin when we
untar the archive. This is bad.

Confirmed bug by
```
# chmod 700 /opt
# cp $archive /opt
# tar xf $archive
# ls -ld /opt
drwxrwxr-x. 6 root root 4096 Oct  6 07:12 /opt
```

Tested fix by
```
# chmod 700 /opt
# cp $archive /opt
# tar xf $archive
# ls -ld /opt
drwx------. 6 root root 4096 Oct  6 11:16 /opt
```

### DIFF
--- a/packaging/bdist_prestoadmin.py
+++ b/packaging/bdist_prestoadmin.py
@@ -146,7 +146,8 @@ class bdist_prestoadmin(Command):
         archive_file = os.path.join(dist_dir, archive_basename)
         self.mkpath(os.path.dirname(archive_file))
         self.make_archive(archive_file, 'bztar',
-                          root_dir=os.path.dirname(build_dir))
+                          root_dir=os.path.dirname(build_dir),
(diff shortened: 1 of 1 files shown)


ebd2:
@petroav: fix for the issue you ran into.
@cawallin, @rschlussel  please review.

cawallin:
Thanks for fixing this!! I thought this was just a tar issue, didn't know we could do anything about it. Looks good.

petroav:
Let me try it out now. Once I confirm that it works for me as well, I'll give you the green light. Thanks for taking a look!

petroav:
Confirmed that it works. Go ahead and push.

PR #70, opened 2015-10-13
Use the dummy presto rpm in product tests when possible

ebd2:
The product tests install presto a large number of times to test various
behaviors of the installation code. In most cases we don't need to actually
install presto, because we don't do anything that requires having presto
installed (e.g. starting presto). For these cases, we can install the dummy
presto rpm instead. Because it's much smaller, presto-admin spends less
time sending the rpm to the nodes in the cluster, and the tests run faster.

### DIFF
--- a/tests/product/base_product_case.py
+++ b/tests/product/base_product_case.py
@@ -41,6 +41,8 @@ PRESTO_VERSION = r'presto-main:.*'
 RETRY_TIMEOUT = 120
 RETRY_INTERVAL = 5
 
+DUMMY_RPM_NAME = 'dummy-rpm.rpm'
+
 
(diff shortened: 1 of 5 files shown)


cawallin:
Looks good % comment

ebd2:
There wasn't a need. The dummy rpm you added already had python as a dependency because you only took out the content and not the other metadata.

akshatnair:
LGTM. Just mention in the commit message what tests were run, as you mentioned earlier (full product tests)

ebd2:
Amended commit message per Akshat, pushed the local branch to github. Closing the pull request manually as the SHA changed so it didn't close automatically.

PR #69, opened 2015-10-09
Add installation for slider itself.

ebd2:
This adds the capability to install and uninstall slider from a node on the
cluster and adds interactive configuration for the minimum config options
needed to install presto and run slider for further invocations.

Product tests are a WIP, I'll add them in another commit on the branch.

Besides the parts directly pertaining to installing slider, I've factored some of the validators out of topology.py into a module of their own under util. This is reflected in a refactoring of the tests for the validators as well. Having modified those files, I also took care of some PEP-8 errors that came up when I ran `make lint`.

I'm not sure I love how the _SLIDER_CONFIG has ended up. I've used this pattern in the past to separate the "interesting" parts of user interactions from the structure of the control flow, and had it come out cleaner. If somebody has a better idea, I'm open to it.

I would appreciate review from @anusudarsan, @rschlussel, and @cawallin. In the meantime, I'm going to start a broader refactoring of the test code to extract the standalone-specific parts out of BaseProductTest.

PR #72, opened 2015-10-19
 Refactor presto/prestoadmin installation out of base_product_case.

ebd2:
In order to support writing product tests for the YARN integration with
slider, we need to refactor the product tests so that we can reuse the
existing docker/configurable cluster logic with slider-based presto
installation and configuration. This change extracts the installation of
presto-admin, presto, and setting the default topology from base_product_case
and moves it to individual modules.

A quick guide to the changes:
In the beginning base_product_case contained the logic for installing presto and presto-admin on a cluster, and for setting the topology. This worked, and code in base_product_case handled setting up clusters with and without presto for various tests.
With the upcoming changes to support YARN integration, we need to use base_product_test as a base class for tests that install presto using slider. Those tests will share the need to set up a cluster and install presto-admin, but the configuration and presto installation will be different. To make it a little easier to do that work I've done the following:
Pulled most of the installation logic out of base_product_case and separated it into modules that install the following:
* prestoadmin_installer: install presto-admin
* topology_installer: install the topology
* presto_installer: install presto

A cluster type is now defined by the set of installers that have to be run against a bare cluster to bring the cluster to the requested state. Each of docker_cluster and configurable_cluster can supply a bare cluster to base_product_case, which is then responsible for running the installers before returning the cluster to the caller.
Additionally, docker_cluster knows how to commit a cluster as an image to use to create clusters again in the future without going through the expensive installation process again. Some of this capability existed before, but we were only using it for a fully-installed presto cluster. In particular, we were recreating the base image every time we created a base cluster. The new behavior is to save an image for reuse for every cluster type, including for the bare cluster.

The tests currently run against a docker cluster. I'll update the PR with any changes needed to run against a configurable_cluster later.

Update:
---

I have the smoke tests running against a vagrant cluster and so far things appear to be OK. I'm cautiously optimistic that there won't be a big change to get configurable cluster fully working.

Also, can I please get a look from @cawallin, @rschlussel.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,13 +40,16 @@ clean-pyc:
 	find . -name '*~' -exec rm -f {} +
 	find . -name '__pycache__' -exec rm -fr {} +
 
+clean-test-containers:
+	 docker ps --format "{{.ID}} {{.Image}}" | awk '/teradatalabs\/pa_test/ { print $$1 }' | xargs docker kill
+
 clean-test:
 	rm -fr .tox/
 	rm -f .coverage
 	rm -fr htmlcov/
 	rm -fr tmp
 	for image in $$(docker images | awk '/teradatalabs\/pa_test/ {print $$3}'); do docker rmi -f $$image ; done
-	echo "Note: The above command is just cleaning up a Docker image that may not exist. If the command fails, it is not a problem."
+	@echo "\n\tYou can kill running containers that caused errors removing images by running \`make clean-test-containers'\n"
 
 lint:
 	flake8 prestoadmin packaging tests
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ clean-test:
 	rm -f .coverage
 	rm -fr htmlcov/
 	rm -fr tmp
-	for image in $(docker images | awk '/teradatalabs\/pa_test/ {print $3}'); do docker rmi -f $image ; done
+	for image in $$(docker images | awk '/teradatalabs\/pa_test/ {print $$3}'); do docker rmi -f $$image ; done
 	echo "Note: The above command is just cleaning up a Docker image that may not exist. If the command fails, it is not a problem."
 
 lint:
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ clean-test:
 	rm -f .coverage
 	rm -fr htmlcov/
 	rm -fr tmp
-	docker rmi -f teradatalabs/centos-presto-test-master teradatalabs/centos-presto-test-slave || true
+	for image in $(docker images | awk '/teradatalabs\/pa_test/ {print $3}'); do docker rmi -f $image ; done
 	echo "Note: The above command is just cleaning up a Docker image that may not exist. If the command fails, it is not a problem."
 
 lint:
--- a/tests/docker_cluster.py
+++ b/tests/docker_cluster.py
@@ -32,16 +32,19 @@ from docker.utils.utils import kwargs_from_env
 from nose.tools import nottest
 
 from prestoadmin import main_dir
+from tests.product.constants import LOCAL_RESOURCES_DIR
 
 INSTALLED_PRESTO_TEST_MASTER_IMAGE = 'teradatalabs/centos-presto-test-master'
 INSTALLED_PRESTO_TEST_SLAVE_IMAGE = 'teradatalabs/centos-presto-test-slave'
 DEFAULT_DOCKER_MOUNT_POINT = '/mnt/presto-admin'
 DEFAULT_LOCAL_MOUNT_POINT = os.path.join(main_dir, 'tmp/docker-pa/')
-LOCAL_RESOURCES_DIR = os.path.join(main_dir, 'tests/product/resources/')
 DIST_DIR = os.path.join(main_dir, 'tmp/installer')
 
 
 class DockerCluster(object):
+    IMAGE_NAME_BASE = os.path.join('teradatalabs', 'pa_test')
+    BARE_CLUSTER_TYPE = 'bare'
+
     """Start/stop/control/query arbitrary clusters of docker containers.
 
     This class is aimed at product test writers to create docker containers
@@ -264,9 +267,8 @@ class DockerCluster(object):
             )
 
     def _ensure_docker_containers_started(self, image):
-        centos_based_images = ['teradatalabs/centos6-ssh-test',
-                               INSTALLED_PRESTO_TEST_MASTER_IMAGE,
-                               INSTALLED_PRESTO_TEST_SLAVE_IMAGE]
+        centos_based_images = ['teradatalabs/centos6-ssh-test']
+
         timeout = 0
         is_host_started = {}
         for host in self.all_hosts():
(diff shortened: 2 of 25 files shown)


Makefile
```diff
@@ -45,7 +45,7 @@ clean-test:
 	rm -f .coverage
 	rm -fr htmlcov/
 	rm -fr tmp
-	docker rmi -f teradatalabs/centos-presto-test-master teradatalabs/centos-presto-test-slave || true
+	for image in $(docker images | awk '/teradatalabs\/pa_test/ {print $3}'); do docker rmi -f $image ; done
```
cawallin: nice!

anusudarsan:
lgtm mod comments. Are you planning to update PR #69 after this?

tests/docker_cluster.py
```diff
@@ -334,67 +336,80 @@ def exec_cmd_on_host(self, host, cmd, raise_error=True, tty=False):
...
-            return presto_cluster
+
+        centos_cluster.start_containers(master_name, slave_name)
+
+        return centos_cluster
 
     @staticmethod
-    def start_base_cluster():
+    def start_existing_images(cluster_type):
+        DC = DockerCluster
```
cawallin: should be lower case. I'm not sure I like this abbreviation; it doesn't fit in with the presto style.

tests/docker_cluster.py
```diff
@@ -334,67 +336,80 @@ def exec_cmd_on_host(self, host, cmd, raise_error=True, tty=False):
...
-            'teradatalabs/centos6-ssh-test',
-            'teradatalabs/centos6-ssh-test'
-        )
+
+        centos_cluster.start_containers(master_name, slave_name)
         return centos_cluster
 
     @staticmethod
-    def _check_for_presto_images():
+    def _check_for_images(master_image_name, slave_image_name):
```
cawallin: In the future, can you separate out the rename changes from the substantial changes? it makes it easier to review

tests/docker_cluster.py
```diff
@@ -334,67 +336,80 @@ def exec_cmd_on_host(self, host, cmd, raise_error=True, tty=False):
...
+            if master_image_name in image['RepoTags'][0]:
                 has_master_image = True
-            if INSTALLED_PRESTO_TEST_SLAVE_IMAGE in image['RepoTags'][0]:
+            if slave_image_name in image['RepoTags'][0]:
                 has_slave_image = True
         return has_master_image and has_slave_image
 
+    def commit_images(self, cluster_type):
+        self.client.commit(self.master,
+                           self._get_master_image_name(cluster_type))
```
cawallin: make sure to run lint, this is a bit suspicious indentation (though it could just be github)

tests/docker_cluster.py
```diff
@@ -419,14 +434,6 @@ def write_content_to_host(self, content, path, host):
     def copy_to_host(self, source_path, dest_host):
         shutil.copy(source_path, self.get_local_mount_dir(dest_host))
 
-    @nottest
-    def clean_up_presto_test_images(self):
```
cawallin: I think it still may be necessary to clean up the test images -- that way if you're running product tests from intellij it deletes the old images.

tests/product/base_product_case.py
```diff
@@ -108,69 +106,55 @@ class BaseProductTestCase(BaseTestCase):
...
         self.cluster = None
+        self.default_keywords = {}
+
+    def _run_installers(self, installers):
+        cluster = self.cluster
+        for installer in installers:
+            dependencies = installer.get_dependencies()
+
+            for dependency in dependencies:
+                dependency.assert_installed(self, cluster.master)
```
cawallin: why does this assume cluster.master?

tests/product/base_product_case.py
```diff
@@ -17,34 +17,32 @@
...
 PRESTO_VERSION = r'presto-main:.*'
 RETRY_TIMEOUT = 120
 RETRY_INTERVAL = 5
 
-DUMMY_RPM_NAME = 'dummy-rpm.rpm'
-
 
 class BaseProductTestCase(BaseTestCase):
+    _cluster_types = {
+        'bare': [],
```
cawallin: can you extract these as constants? e.g. 'bare', 'pa_only', etc

tests/product/presto_installer.py
```diff
@@ -0,0 +1,168 @@
...
+            cluster = testcase.cluster
+
+        try:
+            check_rpm = cluster.exec_cmd_on_host(
+                container, 'rpm -q presto-server-rpm')
+            testcase.assertRegexpMatches(
+                check_rpm, RPM_BASENAME + '\n', msg=msg
+            )
+        except OSError as e:
+            if isinstance(testcase.cluster, DockerCluster):
```
cawallin: we can probably remove this commit, @mattsfuller probably found the underlying cause of presto-admin failures

tests/product/test_control.py
```diff
@@ -142,21 +141,23 @@ def test_server_restart_various_states(self):
             running_host=self.cluster.internal_slaves[0])
 
     def test_start_stop_restart_coordinator_down(self):
-        self.setup_cluster()
-        self.install_presto_admin(self.cluster)
+        installer = PrestoInstaller(self)
```
cawallin: small, but could you rename this to presto_installer so it's more clear?

tests/product/test_server_uninstall.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
```
cawallin: why did you remove this? check if others were also removed

cawallin:
looks good % small comments and running lint

Makefile
```diff
@@ -45,7 +45,7 @@ clean-test:
 	rm -f .coverage
 	rm -fr htmlcov/
 	rm -fr tmp
-	docker rmi -f teradatalabs/centos-presto-test-master teradatalabs/centos-presto-test-slave || true
+	for image in $$(docker images | awk '/teradatalabs\/pa_test/ {print $$3}'); do docker rmi -f $$image ; done
 	echo "Note: The above command is just cleaning up a Docker image that may not exist. If the command fails, it is not a problem."
```
rschlussel: We can get rid of this warning now, right?  Because it's only iterating through images that exist.

tests/docker_cluster.py
ebd2: lint passes, I think this one is github

tests/product/base_product_case.py
ebd2: How about if cluster exposes a get_master() method for installers and then they can pick?

tests/product/test_server_uninstall.py
ebd2: Spurious change. I'll revert it.

tests/docker_cluster.py
ebd2: Let's sync on this. It's not clear to me how the place we're calling this currently is accomplishing cleaning up the images when we run this from intellij

Makefile
ebd2: I've updated the warning to reflect how it might fail now and what to do if it does.

cawallin:
looks good, please squash the feedback commits together

PR #73, opened 2015-10-19
Fix configurable cluster password configuration.

ebd2:
Testing indicated that the user_password was not getting set in time for it
to be picked up by the top-level import in configurable_cluster.py. Moving the
import to the place where user_password is used fixes this problem.


cawallin:
looks good

PR #74, opened 2015-10-20
Correctly parenthesize the tuple in the string formatting expression.

ebd2:
The right-hand argument to the string formatting operator is missing
parentheses, so the second argument is being treated as an argument to
exec_cmd_on_host rather than being included in the formatting operation. This
results in an error because there aren't enough arguments for the format
string.

Testing: re-ran a test that was failing with the string format error. The
test now passes.


cawallin:
looks good

PR #75, opened 2015-10-23
Make stdout/stderr capture optional in BaseTestCase

ebd2:
In order to debug issues in the product tests that we only see on OpenStack,
it would be super helpful to be able to print messages for debugging and see
the output, as we can't log in to any of the docker containers or even debug
the tests themselves. This change makes output capture optional and enables it
for the unit tests that require it.

### DIFF
--- a/tests/base_test_case.py
+++ b/tests/base_test_case.py
@@ -39,8 +39,9 @@ class BaseTestCase(unittest.TestCase):
     old_stderr = sys.__stderr__
     env_vars = None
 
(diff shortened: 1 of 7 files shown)


cawallin:
Aren't you missing the BaseTestCase change?

ebd2:
Nope! I absolutely did not completely forget to include BaseTestChange in the original push to the branch. Nothing to see here :-)

cawallin:
looks good

PR #76, opened 2015-10-23
Improve clean-test and clean-test-containers targets

ebd2:
clean-test tries to clean up all of the docker images that the product tests
create, and if it fails, it should show a helpful message telling the user how
to kill off any running containers. Add a dash to the front of the image
cleanup line so that make ignores the return code and actually shows you the
help if docker couldn't delete an image.

clean-test-containers deletes any running containers that got left behind by
the product tests. Piping a list of containers in to xargs docker kill causes
and error if there aren't any running containers, so I switched it to a loop
like the line that handles deleting images.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,14 +41,14 @@ clean-pyc:
 	find . -name '__pycache__' -exec rm -fr {} +
 
 clean-test-containers:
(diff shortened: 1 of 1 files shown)


cawallin:
looks good

PR #78, opened 2015-10-26
Add test that presto is started as presto user

ebd2:
* Make sure that the user of server's pid is presto
* change run_prestoadmin_script to run_script_from_prestoadmin_dir and
  extend to be able to run on other hosts than master

Testing: make lint, ran test, sandbox with all tests and new RPM

Conflicts:
	tests/product/test_server_install.py

This is a cherry-pick of a change from the release-1.1 branch that addresses product test failures that began when Facebook merged in changes that run presto server as the presto user in the RPM init.d scripts. Facebook cut a new release with this change, and the product tests have been failing as a result.

I've run the full product tests and verified that this fixes the issue and doesn't introduce any new failures.

### DIFF
--- a/tests/product/base_product_case.py
+++ b/tests/product/base_product_case.py
@@ -210,17 +210,21 @@ query.max-memory=50GB\n"""
             raise_error=raise_error
         )
 
-    def run_prestoadmin_script(self, script_contents, **kwargs):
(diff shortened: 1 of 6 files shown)


cawallin:
looks good

PR #79, opened 2015-11-03
Add support for different base images in the product tests

ebd2:
The product tests are shortly going to require that hadoop is actually installed and running on the cluster for the yarn-slider integration. This change factors the base image creation out of DockerCluster. The existing bare image creation code became NoHadoopBareImageProvider.

For Hadoop, we now have an HdpBareImageProvider, which creates bare images that have HDP installed on them. We don't want to recreate this image with any regularity because it takes a while. Instead we rely on a Docker repo:tag existing that we tag into the teradatalabs/pa_test namespace.

Image cleanup is now done by image name instead of image ID so that we don't destroy the preexisting Docker image. It has a different repo:tag, but shares the ID with the tag under the teradatalabs/pa_test namespace.

As bonus cleanup, I added abstract base classes for clusters and installers so that we (hopefully) don't end up going a long way into a test run before discovering that the contract for a cluster or installer has changed and we forgot to update one.

Lastly everywhere that created a cluster now needs to supply a BareImageProvider. Those changes are split out into another commit so as not to be noise among the interesting stuff. This commit will need to be squashed into the substantive changes in the BaseImageProvider commit and the Makefile change before pushing.

The product tests ran nicely this morning with these changes on a docker cluster, and I let a smoke test run for about 10 minutes against a configurable cluster to ensure that the cluster changes hadn't broken anything.

Review by @anusudarsan, @cawallin, and/or @rschlussel would be appreciated.

Thanks!

### DIFF
--- a/tests/bare_image_provider.py
+++ b/tests/bare_image_provider.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Abstract base class for bare image providers.
+
+Bare image providers know how to bring bare docker images into existence for
+the product tests.
+"""
+
+import abc
+
+
+class BareImageProvider:
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
+    def create_bare_images(self, cluster, master_name, slave_name):
+        """Create master and slave images to be tagged with master_name and
+        slave_name, respectivesly."""
+        pass
+
+    @abc.abstractmethod
+    def get_tag_decoration(self):
+        """Returns a string that's prepended to docker image tags for images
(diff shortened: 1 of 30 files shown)


tests/bare_image_provider.py
```diff
@@ -0,0 +1,38 @@
...
+import abc
+
+
+class BareImageProvider:
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
+    def create_bare_images(self, cluster, master_name, slave_name):
+        """Create master and slave images to be tagged with master_name and
+        slave_name, respectivesly."""
```
anusudarsan: respectively typo

tests/product/yarn_slider/slider_installer.py
```diff
@@ -0,0 +1,50 @@
...
+        tsi.upload_config(self.testcase.cluster, conf)
+        slider_path = tsi.copy_slider_dist_to_cluster(self.testcase)
+        tsi.install_slider_package(self.testcase, slider_path)
+
+    def get_keywords(self, *args, **kwargs):
+        return {}
+
+    @staticmethod
+    def assert_installed(testcase):
+        from tests.product.yarn_slider.test_slider_installation import \
```
anusudarsan: move import to top

anusudarsan:
nice! lgtm.

tests/bare_image_provider.py
ebd2: Nice catch. Fixed.

tests/product/yarn_slider/slider_installer.py
ebd2: This breaks an import cycle :-(

rschlussel:
lgtm

PR #80, opened 2015-11-09
Pin the paramiko version to 1.15.3

ebd2:
With the release of paramiko 1.16.0, we've experienced breakage in our tests.
We don't have an explicit dependency on paramiko, so we've never pinned a
version, and just taken what we got when the dependency was resolved
implicitly. Pending further investigation into the breakage, we're making our
dependency on paramiko explicit and pinning it to a known-good version.


ebd2:
I've re-run the failing tests (they pass) and have a smoke test running now.

rschlussel:
looks good

PR #81, opened 2015-11-10
Exit with a non-zero exit code when operations fail.

ebd2:
presto-admin should adhere to the convention that a failure results in a
non-zero exit code. presto-admin runs stuff on potentially many hosts, and
actions can fail on some hosts but not others. This change takes the approach
that in order to exit with an exit code of zero, every action taken against
every host must succeed.

A quick tour of the changes:

Because of the way fabric works, we see failures as SystemExit, which we were previously catching and ignoring. We now stick the exception into the per-host results dictionary. The dictionary maps hosts to exceptions. If none of the hosts have an exception, we exit with an exit code of zero, otherwise we exit with an exit code of one.

In order to propagate failure information correctly, I needed to update a bunch of decorators to return the result of the wrapped function. I wrote most of 'em wrong in the first place, so mea culpa on that one :-)

In the tests, there are a number of places where we were invoking presto-admin and parsing error output. Because pa now exits with a non-zero exit code on error, I had to update the tests so that we don't raise an error when pa exits with a non-zero exit code.

Separated into 2 commits for easy reviewing, will squash before pushing as they need to go in together. If you look at the individual commits, the change to docker_cluster.py in 9491aba is spurious and I undid it in ade1c03


tests/product/test_control.py
```diff
@@ -303,9 +308,10 @@ def assert_simple_start_stop(self, expected_start, expected_stop):
 
     def assert_simple_server_restart(self, expected_output,
                                      running_host='all',
-                                     expected_stop=''):
+                                     expected_stop='', pa_raise_error=True):
```
rschlussel: this looks weirdly formatted.

rschlussel:
looks good

tests/product/test_control.py
ebd2: fixed.

PR #83, opened 2015-11-12
Add presto yarn package install and uninstall

ebd2:
This PR adds the presto-admin functionality for installing and uninstalling the presto-yarn package using slider, and product tests for that.

The changes are mostly straightforward. Installing and uninstalling are much the same as for the rpm package, with the key difference that instead of installing the package using rpm, we install it using slider.

The product tests verify that installation and uninstallation work by checking two things:
1. The package file ends up in the right location on HDFS 
2. Slider recognizes that the package has been installed.

Doing 2. requires running slider as the configured user; this required updating docker-py to support the user= kwarg to exec_create, which in turn required updating how we pass the mem_limit parameter in docker_cluster.py.

In the course of doing this work, I ran into problems with how I was maintaining and updating BaseProductCase's dict of default keywords. I ended up reworking that, with the result that StandalonePrestoInstaller now maintains its notion of dummyness as an instance variable rather than a parameter passed to individual functions. A bunch of minor changes resulted from that. It seems like keyword updates should be less brittle now.


rschlussel:
I haven't really looked at the code yet, but a quick note about docker-py.
1. I don't see where you updated the docker-py version
2. at some point we had intentionally pegged the version of docker-py because the next version was causing some test failures. I don't remember which tests, and I think it may have related to the version of docker we were running, but you should run the suite to make sure it's no longer an issue.

tests/unit/resources/help.txt
```diff
@@ -29,7 +29,9 @@ Commands:
     server stop
     server uninstall
     server upgrade
+    slider install
```
anusudarsan: I thought we wanted this to be called server install ? Because this is not installing slider but setting up Presto on YARN using slider. AND 'slider slider_install' will be just 'slider install'.

prestoadmin/slider/server.py
```diff
@@ -67,3 +74,77 @@ def slider_uninstall():
...
+                   HADOOP_CONF_DIR=conf[HADOOP_CONF]):
+        return sudo(slider_command, user=conf[SLIDER_USER])
+
+
+@task
+@requires_conf
+@task_by_rolename(SLIDER_MASTER)
+def install(presto_yarn_package):
+    """
+    Install the presto-yarn package on the cluster using slider. The
```
anusudarsan: Can you replace slider with Apache Slider? This text is user facing.

prestoadmin/slider/server.py
```diff
@@ -67,3 +74,77 @@ def slider_uninstall():
...
+    slider_command = '%s package --install --package %s --name %s' % \
+                     (get_slider_bin(conf), package_file, conf[APPNAME])
+
+    try:
+        run_slider(slider_command, conf)
+
+        env.conf[PRESTO_PACKAGE] = package_filename
+        store_conf(env.conf, SLIDER_CONFIG_PATH)
+
+        local('unzip %s %s -d %s' %
```
anusudarsan: why are you unzipping here after package --install?

prestoadmin/slider/server.py
```diff
@@ -67,3 +74,77 @@ def slider_uninstall():
...
+    conf = env.conf
+    package_filename = os.path.basename(presto_yarn_package)
+    package_file = os.path.join('/tmp', package_filename)
+
+    result = put(presto_yarn_package, package_file)
+    if result.failed:
+        abort('Failed to send slider application package to %s on host %s' %
+              (package_file, env.host))
+
+    slider_command = '%s package --install --package %s --name %s' % \
```
anusudarsan: rename this as package_install_command

prestoadmin/slider/server.py
```diff
@@ -19,15 +19,21 @@
...
+from prestoadmin.slider.config import requires_conf, store_conf, \
+    DIR, SLIDER_USER, APPNAME, JAVA_HOME, HADOOP_CONF, SLIDER_MASTER, \
+    PRESTO_PACKAGE, SLIDER_CONFIG_PATH, SLIDER_CONFIG_DIR
+
+__all__ = ['slider_install', 'slider_uninstall', 'install', 'uninstall']
 
-from prestoadmin.util.fabricapi import get_host_list, task_by_rolename
 
-__all__ = ['slider_install', 'slider_uninstall']
+SLIDER_PKG_DEFAULT_FILES = ['appConfig-default.json', 'resources-default.json']
```
anusudarsan: can you move the constant  after the import statements?

prestoadmin/slider/server.py
```diff
@@ -67,3 +74,77 @@ def slider_uninstall():
...
+@task
+@requires_conf
+@task_by_rolename(SLIDER_MASTER)
+def uninstall():
+    """
+    Uninstall unregisters the presto application with slider and removes the
+    installed package.
+    """
+    conf = env.conf
+    slider_command = '%s package --delete --name %s' % \
```
anusudarsan: call it package_delete_command

prestoadmin/slider/server.py
```diff
@@ -67,3 +74,77 @@ def slider_uninstall():
...
+        abort('Failed to send slider application package to %s on host %s' %
+              (package_file, env.host))
+
+    slider_command = '%s package --install --package %s --name %s' % \
+                     (get_slider_bin(conf), package_file, conf[APPNAME])
+
+    try:
+        run_slider(slider_command, conf)
+
+        env.conf[PRESTO_PACKAGE] = package_filename
```
anusudarsan: why are you adding this to the env? I cannot find any usages later?

tests/product/yarn_slider/slider_presto_installer.py
```diff
@@ -0,0 +1,115 @@
...
+from prestoadmin.slider.config import APPNAME, SLIDER_USER, HOST, HADOOP_CONF
+from prestoadmin.slider.server import get_slider_bin
+
+from tests.base_installer import BaseInstaller
+from tests.product.prestoadmin_installer import PrestoadminInstaller
+from tests.product.yarn_slider.slider_installer import SliderInstaller
+from tests.product.yarn_slider.test_slider_installation import \
+    TestSliderInstallation
+
+PRESTO_YARN_PACKAGE_GLOB = 'presto-yarn-package*.zip'
```
anusudarsan: are u storing this package locally now?

anusudarsan:
some comments/questions.

tests/unit/resources/help.txt
ebd2: Same as last time; this is going to be updated once mode-switching goes in for both slider_install/uninstall and install/uninstall

prestoadmin/slider/server.py
ebd2: Done

prestoadmin/slider/server.py
ebd2: Done

prestoadmin/slider/server.py
ebd2: We need this to fill in the appConfig.json for server start. No use yet, but that's the next thing on the list.

tests/product/yarn_slider/slider_presto_installer.py
ebd2: I'm caching it the same way we cache presto-server-rpm once we download it.

prestoadmin/slider/server.py
ebd2: Done

prestoadmin/slider/server.py
ebd2: I've replaced the first instance with Apache Slider and subsequent ones with Slider.

prestoadmin/slider/server.py
ebd2: This is where we unzip the default json files into /etc/opt/prestoadmin/slider.

tests/product/base_product_case.py
```diff
@@ -131,12 +131,18 @@ def _run_installers(self, installers):
...
             installer_instance.install()
 
+            self.default_keywords.update(installer_instance.get_keywords())
             cluster.postinstall(installer)
 
     def _apply_post_install_hooks(self, installers):
         for installer in installers:
             self.cluster.postinstall(installer)
 
+    def _update_keywords(self, installers):
```
rschlussel: it looks like this name existed before, but could you call this something more descriptive than keywords.  It's not clear what the keywords are for.

prestoadmin/slider/config.py
```diff
@@ -135,17 +149,23 @@ def write(conf):
 
 
 def get_conf_if_missing():
+    """ Loads the configuration if it hasn't already been loaded. Returns True
+    if the config was loaded for the first time, and False otherwise.
+    """
     with settings(parallel=False):
```
rschlussel: How does this work?  You can't set things to be parallel or serial from within a method because at that point it's already been called parallel-ly or serially.  Setting parallel to false will change the attribute, but it doesn't change the behavior.  Since here it first checks if the confs already been loaded, I imagine any issues would be a race condition that you wouldn't necessarily encounter (and it wouldn't matter much if you reloaded the conf several times), but in interactive mode I feel like it could be an issue pretty easily.

prestoadmin/slider/server.py
rschlussel: Is that to be able to update the configuration later?

ebd2: We need to generate appConfig.json and resources.json files for slider to be able to create the app. We can grab reasonble defaults for most of the content from the default files Anu put in the presto-yarn-package, so I extract them for use later.

tests/product/yarn_slider/slider_presto_installer.py
ebd2: I've added downloading the package from a snapshot in the maven repository.

tests/product/base_product_case.py
ebd2: I've renamed this to _update_replacement_keywords

prestoadmin/slider/config.py
ebd2: Boy did this open a can of worms. It's fixed with the config load changes that went in last week.

tests/product/yarn_slider/slider_presto_installer.py
anusudarsan: make sure you get the latest from there.

.gitignore
```diff
@@ -31,3 +31,6 @@ dist/
 
 #mvn targets
 presto-admin-test/target
+
+# presto yarn package for product tests
+presto-yarn-package.zip
```
anusudarsan: filename has 1.0.0-* version too right?

prestoadmin/yarn_slider/server.py
```diff
@@ -0,0 +1,150 @@
...
+    conf = env.conf
+    package_filename = os.path.basename(presto_yarn_package)
+    package_file = os.path.join('/tmp', package_filename)
+
+    result = put(presto_yarn_package, package_file)
+    if result.failed:
+        abort('Failed to send slider application package to %s on host %s' %
+              (package_file, env.host))
+
+    package_install_command = \
```
anusudarsan: where are you creating the hdfs dir /user/yarn/ and chown it by user yarn ? the package --install would fail without it? https://github.com/prestodb/presto-yarn/blob/master/README.md#preparing-other-slider-specific-configuration has the steps we do manually. looks like hdfs dir creation is needed prior to install?

anusudarsan:
lgtm mod comments

.gitignore
ebd2: If you end up downloading it automatically, it gets this filename:

https://github.com/prestodb/presto-admin/pull/83/files#diff-d7ed4990e37abbc78fb669f18c7cfe1cR116

This follows what we do with the presto-server-rpm.

prestoadmin/yarn_slider/server.py
ebd2: Per offline discussion, I'm going to address this in the future. We can't guarantee that we can actually create the user's home directory because we may not be able to su hdfs to create /user/${app_user} and chown it appropriately. Instead we'll check for the existence and ownership of the home directory and provide a useful error if it doesn't exist or isn't owned by the right user.

prestoadmin/config.py
```diff
@@ -27,18 +27,23 @@
 _LOGGER = logging.getLogger(__name__)
 
 
+def get_conf_from_json(conf_file):
```
rschlussel: why?

tests/product/yarn_slider/test_server_install.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def assert_uninstalled(self):
+        self.assertRaisesRegexp(OSError, 'No such file or directory',
+                                self.installer.assert_installed, self)
+
+    def get_config(self):
+        slider_master = self.installer._get_slider_master(self)
+        conf = self.get_file_content(slider_master, SLIDER_CONFIG_PATH)
+        conf_file = StringIO(conf)
+        return get_conf_from_json(conf_file)
```
rschlussel: oh I see why.  But  once this is not actually a file, it's sort of forced to pretend that it is just so that you can use this method.  You could use json.loads() and operate directly on the string.

prestoadmin/yarn_slider/config.py
```diff
@@ -57,11 +64,11 @@
                     'Connection failed for %%(%s)s@%%(%s)s:%%(%s)d. ' +
                     'Re-enter connection information.'),
 
-    SingleConfigItem(DIR, 'Enter the directory to install slider into on ' +
+    SingleConfigItem(DIR, 'Enter the directory to install slider into on '
                      'the slider master:', '/opt/slider', None),
 
     MultiConfigItem([
-        SingleConfigItem(SLIDER_USER, 'Enter a user name for conducting ' +
+        SingleConfigItem(SLIDER_USER, 'Enter a user name for conducting '
```
rschlussel: too many big words.  Makes it seem unnecessarily complicated.  How about 'Enter a user name for running slider'

rschlussel:
looks good

prestoadmin/config.py
ebd2: Reverted.

tests/product/yarn_slider/test_server_install.py
ebd2: Good call on that one. I've changed this and reverted the changes in prestoadmin/config.py.

prestoadmin/yarn_slider/config.py
ebd2: I kept 'on the slider master', but otherwise I agree.

PR #84, opened 2015-11-12
Refactor coordinator and worker to be classes

rschlussel:
Refactor coordinator and worker configuration code to inherit from an
abstract Node class.  This reduces code duplication

Testing: make test, test_configuration product tests

The changes to coordinator and worker are a bit hard to read in the diff, but basically all that happened is everything got indented to be under the Coordinator or Worker class.  And the get_conf method was moved out to the abstract Node class.

### DIFF
--- a/prestoadmin/coordinator.py
+++ b/prestoadmin/coordinator.py
@@ -22,71 +22,58 @@ import logging
 
 from fabric.api import env
 
-import config
-import presto_conf
-from prestoadmin.presto_conf import validate_presto_conf, get_presto_conf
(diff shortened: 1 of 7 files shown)


rschlussel:
@ebd2 @cawallin

cawallin:
This is just refactoring, right, not fixing the config default stuff? if so, looks good

rschlussel:
yup.  Just refactoring

ebd2:
LGTM. Thanks!

rschlussel:
merged

PR #85, opened 2015-11-13
Clean up docker-py Client object

ebd2:
We're leaking file descriptors like it ain't no thang because we haven't been
close()ing the docker-py Client object in DockerCluster. This change adds
calls to close() so as to not hit the open file limit as we add more tests.

Note for review:
Nobody seems to know why we were calling self.tear_down() at the beginning of start_containers(). It appears to be a no-op; even if we've failed to clean up a previous cluster, the hosts have different unique ids from the ones in the cluster the DockerCluster object represents.

If somebody knows why we are doing this, I'm happy to add a save_client argument to tear_down() and only close the client when it's False.

### DIFF
--- a/tests/docker_cluster.py
+++ b/tests/docker_cluster.py
@@ -137,7 +137,6 @@ class DockerCluster(BaseCluster):
 
     def start_containers(self, master_image, slave_image=None,
                          cmd=None, **kwargs):
-        self.tear_down()
         self._create_host_mount_dirs()
 
         self._create_and_start_containers(master_image, slave_image,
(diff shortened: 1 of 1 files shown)


ebd2:
Testing:
Run a set of product tests, watch /proc/$PID/fd and make sure the number doesn't grow, watch docker ps -a and make sure that number isn't growing either.

tests/docker_cluster.py
```diff
@@ -137,7 +137,6 @@ def _is_image_present_locally(self, image_name, tag):
 
     def start_containers(self, master_image, slave_image=None,
                          cmd=None, **kwargs):
-        self.tear_down()
```
cawallin: yeah, this is left over from a previous implementation

cawallin:
looks good

PR #86, opened 2015-11-16
Add support for JAVA8_HOME config

kkc00:
Adds JAVA8_HOME config support to allowe users to
specify the location of java

### DIFF
--- a/prestoadmin/package.py
+++ b/prestoadmin/package.py
@@ -17,7 +17,7 @@ Module for rpm package deploy and install using presto-admin
 """
 import logging
 
-from fabric.context_managers import settings, hide
+from fabric.context_managers import settings, hide, shell_env
 from fabric.decorators import task, runs_once
 from fabric.operations import sudo, put, os, local
 from fabric.state import env
@@ -93,9 +93,15 @@ def rpm_install(rpm_name):
     if env.nodeps:
         nodeps = '--nodeps '
 
-    ret = sudo('rpm -i %s%s' %
-               (nodeps, os.path.join(constants.REMOTE_PACKAGES_PATH,
-                                     rpm_name)))
+    if 'java8_home' not in env or env.java8_home is None:
+        ret = sudo('rpm -i %s%s' %
+                   (nodeps, os.path.join(constants.REMOTE_PACKAGES_PATH,
+                                         rpm_name)))
+    else:
+        with shell_env(JAVA8_HOME='%s' % env.java8_home):
+            ret = sudo('rpm -i %s%s' %
+                       (nodeps, os.path.join(constants.REMOTE_PACKAGES_PATH,
+                                             rpm_name)))
     if ret.succeeded:
         print("Package installed successfully on: " + env.host)
 
--- a/prestoadmin/topology.py
+++ b/prestoadmin/topology.py
@@ -34,7 +34,8 @@ from prestoadmin.util.validators import validate_username, validate_port, \
 
 __all__ = ['show']
 
-PRESTO_ADMIN_PROPERTIES = ['username', 'port', 'coordinator', 'workers']
+PRESTO_ADMIN_PROPERTIES = ['username', 'port', 'coordinator', 'workers',
+                           'java8_home']
 DEFAULT_PROPERTIES = {'username': 'root',
                       'port': '22',
                       'coordinator': 'localhost',
@@ -138,6 +139,10 @@ def prompt_for_workers():
                                     validate=validate_workers_for_prompt)
 
 
+def validate_java8_home(java8_home):
+    return java8_home
+
+
 def validate_workers_for_prompt(workers):
     return validate_workers(workers.split())
 
@@ -153,6 +158,13 @@ def validate(conf):
         pass
     else:
         validate_username(username)
+
+    try:
+        java8_home = conf['java8_home']
+    except KeyError:
+        pass
+    else:
+        validate_java8_home(java8_home)
 
     try:
         coordinator = conf['coordinator']
@@ -208,6 +220,10 @@ def set_env_from_conf():
 
     env.user = conf['username']
     env.port = conf['port']
+    try:
+        env.java8_home = conf['java8_home']
+    except KeyError:
+        env.java8_home = None
     env.roledefs['coordinator'] = [conf['coordinator']]
     env.roledefs['worker'] = conf['workers']
     env.roledefs['all'] = dedup_list(util.get_coordinator_role() +
--- a/tests/product/test_server_install.py
+++ b/tests/product/test_server_install.py
@@ -171,6 +171,44 @@ query.max-memory=50GB\n"""
                                            '/etc/presto/config.properties',
                                            self.default_workers_config_regex_)
 
+    def test_install_with_java8_home(self, dummy=True):
+        for container in self.cluster.all_hosts():
+            self.cluster.exec_cmd_on_host(container,
+                                          "mv /usr/java/jdk1.8.0_40 /usr/")
+        topology = {"coordinator": "master",
+                    "workers": ["slave1", "slave2", "slave3"],
+                    "java8_home": "/usr/jdk1.8.0_40/jre"}
+        self.upload_topology(topology)
+
+        cmd_output = self.installer.install(dummy)
+        expected = installed_all_hosts_output
+
+        actual = cmd_output.splitlines()
+        self.assertEqual(sorted(expected), sorted(actual))
+
+        for container in self.cluster.all_hosts():
+            self.installer.assert_installed(self, container)
+            self.assert_has_default_config(container)
+            self.assert_has_default_connector(container)
+
+    def test_install_failure_without_java8_home(self, dummy=True):
+        for container in self.cluster.all_hosts():
+            self.cluster.exec_cmd_on_host(container,
+                                          "mv /usr/java/jdk1.8.0_40 /usr/")
+        self.upload_topology()
+        cmd_output = self.installer.install(dummy)
+        actual = cmd_output.splitlines()
+        num_failures = 0
+        for line in enumerate(actual):
+            if str(line).find("Error: Required Java version"
+                              " could not be found") != -1:
+                num_failures += 1
+
+        self.assertEqual(4, num_failures)
+
+        for container in self.cluster.all_hosts():
+            self.installer.assert_uninstalled(container, dummy=True)
+
     @attr('smoketest')
     def test_install(self, dummy=False):
         self.upload_topology()
--- a/tests/product/standalone/presto_installer.py
+++ b/tests/product/standalone/presto_installer.py
@@ -165,9 +165,16 @@ class StandalonePrestoInstaller(BaseInstaller):
                             + os.path.join(cluster.mount_dir, rpm_name)
         )
 
-    def assert_uninstalled(self, container, msg=None):
+    def assert_uninstalled(self, container, dummy=False, msg=None):
+        if dummy:
+            rpm_name = os.path.splitext(DUMMY_RPM_NAME)[0]
+        else:
+            rpm_name = os.path.splittext(self.presto_rpm_filename)[0]
+        failure_msg = 'package %s is not installed' % rpm_name
+        rpm_cmd = 'rpm -q %s' % rpm_name
+
         self.testcase.assertRaisesRegexp(
             OSError,
-            'package presto-server-rpm is not installed',
+            failure_msg,
             self.testcase.cluster.exec_cmd_on_host, container,
-            'rpm -q presto-server-rpm', msg=msg)
+            rpm_cmd, msg=msg)


prestoadmin/package.py
```diff
@@ -92,10 +93,14 @@ def rpm_install(rpm_name):
     nodeps = ''
     if env.nodeps:
         nodeps = '--nodeps '
+   
+    filename = "/%s/.bashrc" % env.user
```
cawallin: I don't like changing the user's .bashrc. Try this: http://stackoverflow.com/questions/8313238/best-way-to-add-an-environment-variable-in-fabric

prestoadmin/package.py
```diff
@@ -92,10 +93,14 @@ def rpm_install(rpm_name):
...
+    with open(filename, "a") as myfile:
+        myfile.write("\nexport JAVA8_HOME=%s" % env.JAVA8_HOME)
 
-    ret = sudo('rpm -i %s%s' %
+    ret = sudo('-E rpm -i %s%s' %
```
cawallin: if you use the above link, you might not need to do -E. check on that.

prestoadmin/topology.py
```diff
@@ -104,9 +106,16 @@ def build_conf_interactive():
     return {'username': prompt_for_username(),
             'port': prompt_for_port(),
             'coordinator': prompt_for_coordinator(),
-            'workers': prompt_for_workers()}
+            'workers': prompt_for_workers(), 
+            'JAVA8_HOME': prompt_for_JAVA8_HOME()}
 
 
+def prompt_for_JAVA8_HOME():
```
cawallin: I don't think it should prompt for JAVA8_HOME actually. it's not super likely the user will have to change this, and we don't want to confuse them. they should still be able to change it via the topology file though

prestoadmin/package.py
kkc00: Changed.

prestoadmin/package.py
kkc00: changed.

prestoadmin/topology.py
kkc00: removed prompt.

kkc00:
Applied changes. Remaining to add tests.

prestoadmin/topology.py
```diff
@@ -138,6 +139,9 @@ def prompt_for_workers():
                                     validate=validate_workers_for_prompt)
 
 
+def validate_JAVA8_HOME(JAVA8_HOME):
```
cawallin: Java8_Home

prestoadmin/package.py
```diff
@@ -92,12 +92,21 @@ def rpm_install(rpm_name):
     nodeps = ''
     if env.nodeps:
         nodeps = '--nodeps '
+   
+    if env.JAVA8_HOME_DEFINED == True:
```
cawallin: You don't need this, just check if JAVA8_HOME is null or not

cawallin: You might need to set JAVA8_HOME to null in the prompt part

cawallin:
Much better, you should look at writing product tests now

prestoadmin/topology.py
kkc00: do you mean renaming the function to validate_Java8_Home() ?

cawallin: actually validate_java8_home, but yes

kkc00: done.

prestoadmin/package.py
kkc00: done

prestoadmin/topology.py
```diff
@@ -208,6 +219,11 @@ def set_env_from_conf():
 
     env.user = conf['username']
     env.port = conf['port']
+    env.JAVA8_HOME = None
```
rschlussel: this statement isn't necessary.

prestoadmin/package.py
```diff
@@ -92,12 +92,20 @@ def rpm_install(rpm_name):
     nodeps = ''
     if env.nodeps:
         nodeps = '--nodeps '
-
-    ret = sudo('rpm -i %s%s' %
-               (nodeps, os.path.join(constants.REMOTE_PACKAGES_PATH,
+   
+    if env.JAVA8_HOME is None:
```
rschlussel: Even though on the command line java8_home is all caps, the variable in the code and config file should be lower case to match other env variables.  We use all caps just for constants, and this is something that gets set.

kkc00:
Added test cases, @cawallin , @rschlussel , @ebd2

prestoadmin/package.py
```diff
@@ -92,12 +92,20 @@ def rpm_install(rpm_name):
...
-
-    ret = sudo('rpm -i %s%s' %
-               (nodeps, os.path.join(constants.REMOTE_PACKAGES_PATH,
+   
+    if env.java8_home is None:
+        ret = sudo('rpm -i %s%s' %
+                   (nodeps, os.path.join(constants.REMOTE_PACKAGES_PATH,
+                                     rpm_name)))
+        if ret.succeeded:
+            print("Package installed successfully on: " + env.host)
```
ebd2: Nit: The check on ret should be factored out of the if block so it isn't duplicated.

prestoadmin/topology.py
```diff
@@ -34,11 +34,11 @@
 
 __all__ = ['show']
 
-PRESTO_ADMIN_PROPERTIES = ['username', 'port', 'coordinator', 'workers']
+PRESTO_ADMIN_PROPERTIES = ['username', 'port', 'coordinator', 'workers', 'JAVA8_HOME']
 DEFAULT_PROPERTIES = {'username': 'root',
                       'port': '22',
                       'coordinator': 'localhost',
-                      'workers': ['localhost']}
+                      'workers': ['localhost'] }
```
ebd2: Have you run make lint? I think this is something it would complain about.

prestoadmin/topology.py
```diff
@@ -104,7 +104,7 @@ def build_conf_interactive():
     return {'username': prompt_for_username(),
             'port': prompt_for_port(),
             'coordinator': prompt_for_coordinator(),
-            'workers': prompt_for_workers()}
+            'workers': prompt_for_workers()}
```
ebd2: I think the trailing space is probably also something lint would pick up.

tests/product/test_server_install.py
```diff
@@ -171,6 +171,25 @@ def assert_installed_with_regex_configs(self, master, slaves):
                                            '/etc/presto/config.properties',
                                            self.default_workers_config_regex_)
 
+    def test_install_with_java8_home(self, dummy=True):
+        for container in self.cluster.all_hosts():
+            self.cluster.exec_cmd_on_host(container, "mv /usr/java/jdk1.8.0_40 /usr/")
+        topology = {"coordinator": "master",
+                        "workers": ["slave1", "slave2", "slave3"],
+                    "JAVA8_HOME": "/usr/jdk1.8.0_40/jre"}
```
ebd2: This indentation looks questionable.

prestoadmin/topology.py
```diff
@@ -34,11 +34,11 @@
 
 __all__ = ['show']
 
-PRESTO_ADMIN_PROPERTIES = ['username', 'port', 'coordinator', 'workers']
+PRESTO_ADMIN_PROPERTIES = ['username', 'port', 'coordinator', 'workers', 'JAVA8_HOME']
```
rschlussel: I think the property name should also be lower case.

rschlussel:
You should run make lint and fix the formatting, but aside from minor comments, this looks good.

tests/product/test_server_install.py
```diff
@@ -171,6 +171,25 @@ def assert_installed_with_regex_configs(self, master, slaves):
                                            '/etc/presto/config.properties',
                                            self.default_workers_config_regex_)
 
+    def test_install_with_java8_home(self, dummy=True):
```
cawallin: can you also add a test for if you move where Java 8 is installed but don't set JAVA8_HOME?

prestoadmin/package.py
kkc00: changed.

prestoadmin/topology.py
kkc00: fixed. ran "make lint" without errors.

prestoadmin/topology.py
kkc00: it is capitalized because in the config.json it is in capital.

tests/product/test_server_install.py
kkc00: Added.

kkc00:
Addressed all the comments. @cawallin , @rschlussel , @ebd2

prestoadmin/topology.py
rschlussel: I mean lower case in config.json too

prestoadmin/topology.py
```diff
@@ -208,6 +220,11 @@ def set_env_from_conf():
 
     env.user = conf['username']
     env.port = conf['port']
+    env.java8_home = None
```
ebd2: This is redundant. env.java8_home is guaranteed to be set in either the try block or the except block.

tests/product/test_server_install.py
```diff
@@ -171,6 +171,45 @@ def assert_installed_with_regex_configs(self, master, slaves):
...
+            self.cluster.exec_cmd_on_host(container,
+                                          "mv /usr/java/jdk1.8.0_40 /usr/")
+        self.upload_topology()
+        cmd_output = self.installer.install(dummy)
+        actual = cmd_output.splitlines()
+        num_failures = 0
+        for line in enumerate(actual):
+            if str(line).find("Error: Required Java version"
+                              " could not be found") != -1:
+                num_failures += 1
```
ebd2: We should also assert that the dummy rpm wasn't installed on any of the hosts.

tests/product/test_server_install.py
```diff
@@ -171,6 +171,45 @@ def assert_installed_with_regex_configs(self, master, slaves):
...
+
+        test_status = "fail"
+        if num_failures == 4:
+            test_status = "pass"
+
+        self.assertEqual("pass", test_status)
```
ebd2: 207 to 211 should be replaced with self.assertEqual(num_failures, 4)

tests/product/test_server_install.py
kkc00: Isn't it sufficient ? This is checking that the java version error shows up for 4 times (master + 3 slaves), which implies that the rpm was not installed on any of the nodes.

ebd2: Honestly, we should be checking for the effects we care about as much as possible rather than looking at output all the time.

The check that the package isn't actually installed is the important check of the behavior of the system under the test conditions. I.e. that the package wasn't installed when Java can't be found.

The check on the output is a brittle check that we think we failed for the right reason. If the output or the behavior of the rpm changes for any reason, just checking the outputs can give us misleading results.

prestoadmin/topology.py
cawallin: Also, please don't set it if the java8_home keyword isn't found. this is what we were talking about at lunch

prestoadmin/package.py
```diff
@@ -93,9 +93,15 @@ def rpm_install(rpm_name):
     if env.nodeps:
         nodeps = '--nodeps '
 
-    ret = sudo('rpm -i %s%s' %
-               (nodeps, os.path.join(constants.REMOTE_PACKAGES_PATH,
-                                     rpm_name)))
+    if env.java8_home is None:
```
cawallin: you can check instead if the java8_home key is in the dictionary (see http://stackoverflow.com/questions/1602934/check-if-a-given-key-already-exists-in-a-dictionary)

tests/unit/test_package.py
```diff
@@ -37,6 +37,7 @@ def test_deploy_is_called(self, mock_put, mock_sudo):
     def test_rpm_install(self, mock_sudo):
         env.host = 'any_host'
         env.nodeps = False
+        env.java8_home = None
```
cawallin: you will probably be able to get rid of these changes with the above suggested change

tests/product/test_server_install.py
kkc00: done.

tests/product/test_server_install.py
kkc00: done.

prestoadmin/package.py
kkc00: done

tests/unit/test_package.py
kkc00: done

prestoadmin/topology.py
kkc00: done

kkc00:
Applied all the changes. @cawallin @rschlussel @ebd2

prestoadmin/package.py
```diff
@@ -93,9 +93,18 @@ def rpm_install(rpm_name):
     if env.nodeps:
         nodeps = '--nodeps '
 
-    ret = sudo('rpm -i %s%s' %
-               (nodeps, os.path.join(constants.REMOTE_PACKAGES_PATH,
-                                     rpm_name)))
+    if 'java8_home' in env:
+        print "java8_home in env"
```
ebd2: Debugging print?

tests/product/test_server_install.py
```diff
@@ -171,6 +171,50 @@ def assert_installed_with_regex_configs(self, master, slaves):
...
+        self.assertEqual(4, num_failures)
+
+        no_rpm_installs = 0
+        for container in self.cluster.all_hosts():
+            output = self.cluster.exec_cmd_on_host(container, "rpm -qa")
+            outputrpms = output.splitlines()
+            if "dummy-rpm.rpm" not in outputrpms:
+                no_rpm_installs += 1
+
+        self.assertEquals(4, no_rpm_installs)
```
ebd2: There is an assert_unininstalled() method in PrestoInstaller that you should use instead of this.

kkc00: I am not sure how it is used, do you have suggestions for how to do in this case ? Are you saying that the whole uninstall test function should be removed ? or any specific part ?

prestoadmin/package.py
kkc00: yes, will remove.

kkc00: done.

tests/product/test_server_install.py
kkc00: Changed, used assert_uninstalled(), also modified assert_uninstalled() to support dummy rpm uninstall.

kkc00:
Applied all changes, @cawallin , @rschlussel , @ebd2

kkc00:
@cawallin , squashed all the commits and did push --force to the latest commit (after rebase).

Issue #87, opened 2015-11-16
Possible error in automatically filling out config.json


petroav:
During server start presto-admin gives the following error message:

```
Fatal error: coercing to Unicode: need string or buffer, int found

Underlying exception:
    coercing to Unicode: need string or buffer, int found

Aborting.
```

Putting double quotes around the port number in `/etc/opt/prestoadmin/config.json` fixed the issue.

rschlussel:
Is this when the config was set interactively?

petroav:
Yes.

rschlussel:
Fixed by https://github.com/prestodb/presto-admin/issues/88

PR #88, opened 2015-11-17
Return string from validate port

rschlussel:
Validate port was returning an int, which was causing casting problems.
Change to return a string

Testing: unit tests, update product test, manual verification

### DIFF
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -183,7 +183,7 @@ def service(control=None):
 
 
 def check_status_for_control_commands():
-    client = PrestoClient(env.host, env.port)
+    client = PrestoClient(env.host, env.user, env.port)
     print('Waiting to make sure we can connect to the Presto server on %s, '
           'please wait. This check will time out after %d minutes if the '
           'server does not respond.'


rschlussel:
@petroav @cawallin .  This is for https://github.com/prestodb/presto-admin/issues/87

tests/product/test_server_install.py
```diff
@@ -351,6 +351,10 @@ def test_install_interactive_with_hostnames(self):
             self.assert_has_default_connector(container)
             self.assert_has_jmx_connector(container)
 
+        # make sure that we don't get an error when using the written out
+        # configs
+        self.run_prestoadmin('server start')
```
cawallin: this is the real rpm not the dummy one?

prestoadmin/util/validators.py
```diff
@@ -31,18 +31,18 @@ def validate_username(username):
 
 
 def validate_port(port):
+    if not isinstance(port, basestring):
```
cawallin: I think we should accept either a string or an integer, we want to avoid times when the user is unexpectedly told things don't work

tests/product/test_server_install.py
cawallin: we should also make sure there isn't an error (though maybe with @ebd2 's change it'll fail out immediately?)

rschlussel: yup.  It fails if there's an error

rschlussel:
I tracked down that the error was coming from the port being passed in instead of the user.  The error came when it tried logging the user information because the user was never expected to be an int, so it wasn't cast as a string first.

cawallin:
do you need to update unit testing? other than that looks good

anusudarsan:
lgtm

rschlussel:
No the unit tests all mock this out.  I could add a unit test that just checks that for this function prestoclient is called with these arguments.  That seems awfully specific though.

cawallin:
I guess that's fine then... I hate python mocks!

rschlussel:
merged

PR #92, opened 2015-11-25
Deploy user supplied confs without adding defaults

rschlussel:
Changes behavior of `configuration deploy` to deploy any user defined
config files as is, without adding in default values for undefined
"required properties".  If no configuration file is present for
'node.properties', 'config.properties', or 'jvm.config', presto-admin
will deploy a default configuration file.

This change makes it possible for a user to remove properties that
are no longer valid in presto (for example if the user has upgraded to a
different presto version where a property that had been required now no
longer exists).

It also means that to change some config properties to non-default
values, the user must write out the whole configuration file

Testing: make test-all

@ebd2 @anusudarsan  could you please review?
@brian-rickman fyi, and if you have any comments about usability

### DIFF
--- a/prestoadmin/config.py
+++ b/prestoadmin/config.py
@@ -71,6 +71,28 @@ def json_to_string(conf):
     return json.dumps(conf, indent=4, separators=(',', ':'))
 
 
+def write_conf_to_file(conf, path):
+    # Note: this function expects conf to be flat
+    # either a dict for .properties file or a list for .config
+    ext = os.path.splitext(path)[1]
+    if ext == ".properties":
+        write_properties_file(conf, path)
+    elif ext == ".config":
+        write_config_file(conf, path)
+
+
+def write_properties_file(conf, path):
+    output = ''
+    for key, value in conf.iteritems():
+        output += '%s=%s\n' % (key, value)
+    write(output, path)
+
+
+def write_config_file(conf, path):
+    output = '\n'.join(conf)
+    write(output, path)
+
+
 def write(output, path):
     conf_directory = os.path.dirname(path)
     try:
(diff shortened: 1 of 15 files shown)


prestoadmin/config.py
```diff
@@ -71,6 +71,26 @@ def json_to_string(conf):
     return json.dumps(conf, indent=4, separators=(',', ':'))
 
 
+def write_conf_to_file(value, path):
```
ebd2: I propose rename 'value' to conf, as it isn't a single value.

prestoadmin/node.py
```diff
@@ -40,21 +41,29 @@ def get_conf(self):
...
 
-        defaults = self.build_defaults()
-        config.fill_defaults(conf, defaults)
         self.validate(conf)
         return conf
 
     @abstractmethod
     def _get_conf_dir(self):
-        pass
+        raise NotImplementedError()
```
ebd2: To further my python knowledge, why the change here? You shouldn't be able to instantiate a Node subclass that doesn't implement this anyway, right?

prestoadmin/config.py
ebd2: Can we please get a comment that notes that this only works for flat configs so that when somebody tries to use it for something hierarchical, they know that they need to add functionality?

ebd2:
This looks generally good to me. I'll wait to push until after you do and I'll sort out any merge issues that arise. My money is on there being several.

prestoadmin/node.py
rschlussel: not needed.  I'm switching it back.  I think I switched it because pycharm gives a type check warning, but it's not a correct warning..

anusudarsan:
lgtm

PR #93, opened 2015-11-25
Load task configuration by callback

ebd2:
The presto-admin configuration should be loaded based on what the task the user is invoking requires to run. This PR does that by doing the following:
1. Unifying (somewhat) configuration between slider and standalone clusters so that both configurations share a common base class that knows how to load their config and whether or not the config has been loaded.
2. Adding a decorator for tasks, @requires_config, that takes a configuration class as an argument. The decorator adds an attribute, pa_config_callback, to the task function that main.py knows to look for and call before running the tasks. At the time the task is run, the decorator verifies that the config has been loaded. If the config hasn't been loaded, it's an error.

That sounds simple enough, but things got interesting along the way.
1. Config load is intertwined with command line argument parsing.
2. Our handling of config defaults is mixed in with this as well.
3. The situation with env.port is "interesting". I think I take much of the blame for changing it to an int a while back. I got to the bottom of it and documented it in test_main.py. TL;DR we should probably change it back to a string at some point because that's what fabric defaults it to.
4. There were a number of tasks that implicitly, but not explicitly required that the topology was loaded.
5. Interactive config was triggered by certain tasks being executed from a limbo state where the config file was absent at the point where we validated command line arguments.

Things have changed a bit:
1. Config load is still intertwined with command line argument parsing, but both parts have moved to after we know what tasks we're going to invoke, and thus know what config to load.
2. All tasks now explicitly require config.
3. Interactive config is triggered before tasks get kicked off by the absence of a config file. We no longer kick of tasks in a weird limbo state.
4. I've removed the tests that explicitly tested what happens when tasks were run in limbo.

### DIFF
--- a/prestoadmin/main.py
+++ b/prestoadmin/main.py
@@ -652,7 +652,8 @@ def _set_arbitrary_env_vars(non_default_options):
 
 
 def validate_hosts(cli_hosts, config_path):
-    # If there's no config file to validate against, don't.
+    # If there's no config file to validate against, don't. This would happen
+    # in the case of a task that doesn't define a callback that loads config.
     if config_path is None:
         return
 
@@ -668,8 +669,8 @@ def validate_hosts(cli_hosts, config_path):
         raise ConfigurationError(
             'Hosts cannot be defined with --hosts/-H when no hosts are listed '
             'in the configuration file %s. Correct the configuration file or '
-            'go through the interactive configuration process by running the '
-            'command again without the --hosts or -H option.' % config_path)
+            'run the command again without the --hosts or -H option.' %
+            config_path)
 
 
 def _update_env(default_options, non_default_options, load_config_callback):
@@ -682,6 +683,8 @@ def _update_env(default_options, non_default_options, load_config_callback):
     else:
         config_path = None
 
+    # Save env.hosts from the config into another env variable for validation.
+    # _set_arbitrary_env_vars will overwrite it if --set hosts=... is present.
     if state.env.hosts:
         state.env.conf_hosts = state.env.hosts
 
@@ -689,7 +692,9 @@ def _update_env(default_options, non_default_options, load_config_callback):
 
     if isinstance(state.env.hosts, basestring):
         # Take advantage of the fact that if there was a generic --set option
-        # for hosts, it's still an unsplit, comma separated string.
+        # for hosts, it's still an unsplit, comma separated string rather than
+        # a list, which is what it would be after loading hosts from a config
+        # file.
         validate_hosts(state.env.hosts, config_path)
 
     # Go back through and add the non-default values (e.g. the values that
--- a/prestoadmin/main.py
+++ b/prestoadmin/main.py
@@ -653,29 +653,47 @@ def _set_arbitrary_env_vars(non_default_options):
     return Values(non_default_options_dict)
 
 
+def validate_hosts(cli_hosts, config_path):
+    # At this point, state.env.conf_hosts contains the hosts that we loaded
+    # from the configuration, if any.
+    cli_host_set = set(cli_hosts.split(','))
+    if 'conf_hosts' in state.env:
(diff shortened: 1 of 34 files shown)


ebd2:
@rschlussel, @cawallin, @anusudarsan Sorry for the giant PR. Most of the substantive changes in prestoadmin are in main.py, topology.py, standalone/config.py, and util/baseconfig.py.

Most of the substantive changes to the tests are in unit/test_main.py. A few tests in unit/test_topology.py have been removed, and tests that test similar functionality have been created in unit/util/test_base_config.py.

ebd2:
Testing done:
unit tests, product tests, lint all pass. I've done some manual sanity checking. If there's anything specific anybody wants me to check, please let me know.

prestoadmin/main.py
```diff
@@ -627,7 +625,7 @@ def _to_boolean(string):
 
 def _set_arbitrary_env_vars(non_default_options):
```
rschlussel: while you're here, could you rename this to _set_non_default_env_vars

tests/unit/test_main.py
```diff
@@ -386,11 +388,112 @@ def test_nodeps_check(self):
...
 
+    def test_get_default_options(self):
+        options = Values({'k1': 'dv1', 'k2': 'dv2'})
+        non_default_options = Values({'k2': 'V2', 'k3': 'V3'})
+        default_options = main.get_default_options(options,
+                                                   non_default_options)
+        self.assertEqual(default_options, Values({'k1': 'dv1'}))
+
+    #
+    # The env.port situation is currently a special kind of hell. There are a
```
rschlussel: So this means that when the port is passed in on the cli it doesn't get validated.  If we validated the port and set the value for that, then it would be consistent.

tests/product/test_server_install.py
```diff
@@ -191,12 +191,12 @@ def test_install_with_java8_home(self, dummy=True):
...
             self.assert_has_default_connector(container)
 
-    def test_install_failure_without_java8_home(self, dummy=True):
+    def test_install_failure_without_java8_home(self):
         for container in self.cluster.all_hosts():
             self.cluster.exec_cmd_on_host(container,
                                           "mv /usr/java/jdk1.8.0_40 /usr/")
         self.upload_topology()
-        cmd_output = self.installer.install(dummy)
+        cmd_output = self.installer.install(dummy=True, pa_raise_error=False)
```
rschlussel: Why did this need to change?

prestoadmin/util/exception.py
```diff
@@ -22,6 +22,8 @@
 import traceback
 
 
+# Beware the nuances of pickling Exceptions:
```
rschlussel: nice

tests/unit/test_main.py
ebd2: Yup. That is precisely the case. I think the next logical step to config is to move all of the validation to validating the env to somewhere after _update_env finishes and before we kick off tasks. I feel like that's probably a task for after the release though :-)

As a side-effect of doing that work, we can ensure that env.port is set to a consistent type regardless of where it came from.

rschlussel:
looks good

tests/product/test_server_install.py
ebd2: The way server install used to work masked the error because it re-executed the task and (presumably) the return value of that wasn't propagated all the way back to main.py where exit_code would notice.

Since server install no longer execute()s deploy_install, errors show up as non-zero exit codes, which raise OSErrors if you don't add this.

rschlussel: oh got it.  Why is package install still runs_once and calling execute(), then?

ebd2: Fixed package install and slider slider_install. Re-ran tests.product.test_package_install and tests.product.yarn_slider.

anusudarsan:
lgtm

rschlussel:
thanks.  lgtm

PR #94, opened 2015-11-30
Change check_if_valid_rpm to be run only once.

mattsfuller:
Make check_if_valid_rpm to only be run once. This required moving the call from deploy_action to install in package.py. install was annoted with @runs_once and now relies on the execute to run in the install in parallel on the cluster.

The motivation behind this change is to hopefully fix an intermittent failure during parallel calls to rpmdb on the same host. In general it makes sense to not run the same check_if_valid function more than once.

### DIFF
--- a/prestoadmin/package.py
+++ b/prestoadmin/package.py
@@ -18,21 +18,23 @@ Module for rpm package deploy and install using presto-admin
 import logging
 
 from fabric.context_managers import settings, hide, shell_env
-from fabric.decorators import task
+from fabric.decorators import task, runs_once
 from fabric.operations import sudo, put, os, local
 from fabric.state import env
+from fabric.tasks import execute
 from fabric.utils import abort
 
 from prestoadmin.util import constants
 from prestoadmin.standalone.config import StandaloneConfig
 from prestoadmin.util.base_config import requires_config
-
+from prestoadmin.util.fabricapi import get_host_list
 
 _LOGGER = logging.getLogger(__name__)
 __all__ = ['install']
 
 
 @task
+@runs_once
 @requires_config(StandaloneConfig)
 def install(local_path):
     """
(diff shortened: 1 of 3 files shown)


mattsfuller:
@cawallin @ebd2

prestoadmin/package.py
```diff
@@ -18,21 +18,23 @@
...
+from prestoadmin.util.fabricapi import get_host_list
 
 _LOGGER = logging.getLogger(__name__)
 __all__ = ['install']
 
 
 @task
+@runs_once
 @requires_config(StandaloneConfig)
 def install(local_path):
```
cawallin: I think you probably also need to add check_if_valid_rpm to server.py's install method.

cawallin:
lgtm % updating server.py also

prestoadmin/package.py
ebd2: Good catch. We need a similar change to server.py's install method to be @runs_once and then execute the remaining steps. See the diff for 82b5cb9b622a94bdc3da9d8ff6429e1c05e545e2 for how it used to be structured.

ebd2:
LGTM except for the server.py stuff.

mattsfuller:
@ebd2 @cawallin 
AC. Does this look good?

prestoadmin/server.py
```diff
@@ -81,10 +82,13 @@ def install(local_path):
     Parameters:
         local_path - Absolute path to the presto rpm to be installed
     """
+    package.check_if_valid_rpm(local_path)
+    execute(deploy_install_configure, local_path, hosts=get_host_list())
```
ebd2: This should be return execute(...

ebd2:
Have you tested this manually and checked the exit code when you do a server install or package install? You may need to incorporate the change from here to ensure that the exit code is non-zero:
https://github.com/prestodb/presto-admin/pull/83/files#diff-0739cfee68158551e8acd1344153fd03R852

mattsfuller:
I have not done that...will do.

mattsfuller:
A non zero exit code is returned
@ebd2 -- I don't have time to address your suggestion until next week. Can I push since you have a PR for your suggested modification anyway?

cawallin:
I think that'd be fine, otherwise this looks good.

PR #95, opened 2015-12-04
do not encourage add custom jars to default connectors in presto-admi…

arhimondr:
…n docs examples


cawallin:
looks good, thanks!

PR #96, opened 2015-12-04
Fix server status for presto >= 0.128

ebd2:
The release of presto 0.128 changed the name of a column we were querying in the implementation of presto-admin server-status.
Fixing the issue involves running a different query against presto depending on the version.
Ensuring consistent output regardless of version requires processing the results of the query differently, again depending on the version.

As part of the work to fix this, I added in general tooling to handle version-based issues. This isn't the first time we've needed to change the behavior of something based on the version, and it surely won't be the last.

I plan to leave this as two separate commits, as the version functionality is independently useful.

### DIFF
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -18,6 +18,7 @@ using presto-admin
 """
 import logging
 import re
+import sys
 
 from fabric.api import task, sudo, env, quiet
 from fabric.context_managers import settings, hide
@@ -30,6 +31,8 @@ from prestoadmin import configure_cmds
 from prestoadmin import connector
 from prestoadmin import package
 from prestoadmin.util.constants import REMOTE_PRESTO_LOG_DIR
+from prestoadmin.util.version_util import VersionRange, VersionRangeList, \
+    split_version, strip_tag
 from prestoadmin.prestoclient import PrestoClient
 from prestoadmin.standalone.config import StandaloneConfig
 from prestoadmin.util.base_config import requires_config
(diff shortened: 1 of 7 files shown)


ebd2:
Testing:
* tests.unit
* tests.product.test_status against 0.115t, 0.124, 0.128

prestoadmin/server.py
```diff
@@ -510,10 +536,12 @@ def get_status_from_coordinator():
         elif not is_running:
             print('\tNo information available')
         else:
+            version_string = get_presto_version()
```
cawallin: can you update check_presto_version() to use this version parsing also?

prestoadmin/util/version_util.py
```diff
@@ -0,0 +1,189 @@
...
+"""
+
+import re
+
+TD_VERSION = re.compile(r'^\d+t$')
+ANCIENT_TAGGED_VERSION = re.compile(r'^\d+t?-[A-Z]+$')
+
+
+def split_version(version_string):
+    return version_string.strip().split('.')
```
maciejgrzybek: Why not inline this?

prestoadmin/util/version_util.py
```diff
@@ -0,0 +1,189 @@
...
+        self.versioned_thing = versioned_thing
+
+    def __str__(self):
+        return '[%s, %s) -> %s' % (
+            '.'.join([str(c) for c in self.min_version]),
+            '.'.join([str(c) for c in self.max_version]),
+            self.versioned_thing)
+
+    @staticmethod
+    def strip_td_suffix(other):
```
maciejgrzybek: `other` is not very meaningful I believe.

prestoadmin/util/version_util.py
```diff
@@ -0,0 +1,189 @@
...
+                self.pad_tuple(other, l, 0))
+
+    def __contains__(self, other):
+        other = self.strip_td_suffix(other)
+        other = tuple([int(component) for component in other])
+
+        min_pad, max_pad, o_pad = self.zero_pad(other)
+        return min_pad <= o_pad and o_pad < max_pad
+
+    def is_continuous(self, next):
```
cawallin: why isn't the next argument used?

prestoadmin/util/version_util.py
```diff
@@ -0,0 +1,189 @@
...
+    :param version: something that can be sliced
+    :return: a tuple containing only integer components, except for possibly
+             the last one, which will be a string iff it's an integer followed
+             by the letter 't'
+    """
+    is_teradata = False
+    is_ancient = False
+
+    result = list(version[:])
+    while True:
```
cawallin: I think a for loop would be clearer here than while True

prestoadmin/util/version_util.py
```diff
@@ -0,0 +1,189 @@
...
+        return tuple(result)
+
+    def zero_pad(self, other):
+        """
+        Pad out min_version, max_version, and other with zeroes to the length
+        of the longest of the three. This allows subsequent comparisons to work
+        as expected when tuples are of unequal length.
+        Returns a tuple of tuples padded out to the same length
+        """
+        l = max(len(self.min_version), len(self.max_version), len(other))
```
cawallin: Do you need to pad it? a quick test suggests that maybe Python tuples are smart enough to handle this?
>>> (0,1) < (0, 1, 2)
True
>>> (0, 1, 2) < (0, 1)
False
>>> (0, 1, 3) < (0, 2)
True

prestoadmin/util/version_util.py
cawallin: I still think a regex would be clearer (because it helps you understand what the version string is supposed to be), but I see the value of longer version string potential. maybe we could compromise and allow for there to be another value (e.g. 0.115t.1) but only 1? I don't think our version strings would get more nested than that.

cawallin:
A couple of comments

prestoadmin/util/version_util.py
ebd2: It's needed for correct comparisons when the tuples are the same modulo trailing zeroes:
>>> (1,) == (1, 0, 0)
False
>>> (1,) < (1, 0, 0)
True
>>> (1,) > (1, 0, 0)
False

cawallin: Is that a valid version string/range?

cawallin: Though probably best to keep it anyway, actually.

prestoadmin/util/version_util.py
ebd2: Because if it ever changes, we'll have n places to change the code instead of 1.

prestoadmin/util/version_util.py
ebd2: I'll comment this more clearly per offline discussion.

cawallin:
ok, looks good

PR #97, opened 2015-12-09
Add make target to produce both online and offline installers

mattsfuller:
There is no target to do both. And each target for the offline and online do cleans making it impossible to run both and keep both artifacts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,6 +89,8 @@ release: clean
 	python setup.py sdist upload -r pypi_internal
 	python setup.py bdist_wheel upload -r pypi_internal
 
+release-builds: clean-build clean-pyc dist dist-online
+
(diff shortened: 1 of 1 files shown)


cawallin:
looks good

rschlussel:
lgtm

Makefile
```diff
@@ -89,6 +89,11 @@ release: clean
 	python setup.py sdist upload -r pypi_internal
 	python setup.py bdist_wheel upload -r pypi_internal
 
+release-builds: clean-build clean-pyc
```
ebd2: Have you tried the following:
release-builds: clean-build clean-pyc dist dist-online
If that works, we don't need to define the dist build process a second time, if not, we should figure out why and fix it.

PR #98, opened 2015-12-09
Update java8_home test because of rpm update

rschlussel:
Updates java8_home test to move java8 to a location other than /usr
because now the rpm will check /usr/bin/java for java8

Testing: ran tests

@cawallin @ebd2


ebd2:
This looks good, but by my reading of https://github.com/Teradata/presto/pull/71 this isn't needed because java 8 ends up in /usr/jdk1.8.blah-blah-blah, which isn't in the list of places we check. Because that isn't on the list of places we check, we shouldn't have any false positives (pass when java 8 has been moved, but JAVA8_HOME isn't set correctly)

rschlussel:
oh you're right.  I'll close this

PR #99, opened 2015-12-10
Basic slider start and stop

ebd2:
In the interest of getting early feedback on what's going to turn into a lengthy PR, here's an implementation of the basic slider start and stop functionality.

Missing things:
- [x] Create the presto data directory on every node in the cluster.
- [x] Which requires querying the ResourceManager's REST API for the list of nodes in the cluster.
- [ ] Product tests for start/stop.
- [x] Compute yarn.memory in resources.json from the jvm heap size specified by appConfig.json
- [x] Doc strings for new tasks in prestoadmin/yarn_slider/server.py
- [x] Comments generally in the new code.
- [x] Review the updated stuff in appConfig-default.json and update the transformations if needed.

Just as a heads-up, the hadoop config parsing is super rough and ready. I spent about 15 minutes looking for something that would do the work. Not having found anything, I rolled my own. It's written with the assumption that the configurations are basically valid (otherwise the hadoop cluster would be screwed up) and the error checking is limited to sanity checking. I'll put in a comment to this effect as well.

Sorry for the totally screwed up commit history. Rebasing produced a lot of conflicts. I'll squash this all together at the end, but for now I'm going to work on the missing functionality.

Would appreciate a look from @anusudarsan for functionality and a general look for sanity from @rschlussel or @cawallin, depending on who has time/motivation.

Like the previous stuff, the commands are currently exposed as pa slider start/stop/etc. This will remain true until we get mode selection in, which is probably the next thing to do.


prestoadmin/yarn_slider/server.py
```diff
@@ -117,7 +124,7 @@ def install(presto_yarn_package):
 
     package_install_command = \
         '%s package --install --package %s --name %s' % \
-        (get_slider_bin(conf), package_file, conf[APPNAME])
+        (get_slider_bin(conf), package_file, SLIDER_PKG_NAME)
```
anusudarsan: should this also have --replacepkg (just in case the package already exists? I think it is ok to assume here that replacing a package is OK.

prestoadmin/yarn_slider/server.py
```diff
@@ -151,5 +158,97 @@ def uninstall():
...
+            slider_cmd = '%s %s %s --template %s --resources %s' % (
+                get_slider_bin(env.conf), command, env.conf[APPNAME],
+                remote_appConfig, remote_resources)
+            return run_slider(slider_cmd, env.conf)
+
+
+@task
+@requires_config(SliderConfig, AppConfigJson, ResourcesJson)
+@task_by_rolename(SLIDER_MASTER)
+def create():
```
anusudarsan: docstring

prestoadmin/yarn_slider/server.py
```diff
@@ -151,5 +158,97 @@ def uninstall():
...
+def _build_or_create(command=BUILD):
+    appConfig = AppConfigJson()
+    resources = ResourcesJson()
+
+    with remote_temp_file(appConfig.get_config_path(), '/tmp') as \
+            remote_appConfig:
+        with remote_temp_file(resources.get_config_path(), '/tmp') as \
+                remote_resources:
+            slider_cmd = '%s %s %s --template %s --resources %s' % (
+                get_slider_bin(env.conf), command, env.conf[APPNAME],
```
anusudarsan: should the APPNAME a input argument to the task create()?

prestoadmin/yarn_slider/server.py
```diff
@@ -151,5 +158,97 @@ def uninstall():
...
+# app doesn't exist and we have to fall back to creating it, it's too late to
+# prompt the user for anything we don't have a default for when we call
+# create().
+@task
+@requires_config(SliderConfig, AppConfigJson, ResourcesJson)
+@task_by_rolename(SLIDER_MASTER)
+def start():
+    appname = env.conf[APPNAME]
+    start_result = start_app()
+    if start_result.return_code == EXIT_SUCCESS:
```
anusudarsan: hmm..so if we run start/create() and the bin/slider command succeeds, but later the YARN application created fails to actually launch the process - this will give a wrong impression to the user than the application is up and running, correct? Is there any other way we can handle that? 

Probably we can leave it to the user to do a status check later to see the application status. Wondering if there should be a status check call here after create/start. Like in standalone mode, I think we do call status() after a server start?

prestoadmin/yarn_slider/slider_application_configs.py
```diff
@@ -0,0 +1,95 @@
...
+            APP_CONFIG_TRANSFORMATIONS)
+
+
+def _prompt_worker_instances(kpath, value):
+    return prompt(
+        "Enter the number of worker nodes to create", default=value)
+
+
+RESOURCES_TRANSFORMATIONS = {
+    as_key('components', 'COORDINATOR', 'yarn.label.expression'): purge,
```
anusudarsan: is there anything from the YARN label setup steps we can automate here? Like call to yarn commands to setup labels on nodes. Probably we can have the user do it for phase2 and add those things later

prestoadmin/yarn_slider/server.py
ebd2: I think you'll get useful output from the call to slider create. We aren't hiding it from the user.

prestoadmin/yarn_slider/slider_application_configs.py
```diff
@@ -0,0 +1,95 @@
...
+APP_CONFIG_TRANSFORMATIONS = {
+    as_key('global', 'site.global.app_user'):
+        lambda kpath, value: env.conf[SLIDER_USER],
+    as_key('global', 'site.global.data_dir'): _prompt_data_dir,
+    as_key('global', 'application.def'): _application_def,
+    as_key('global', 'java.home'):
+        lambda kpath, value: env.conf[JAVA_HOME],
+    as_key('global', 'site.global.singlenode'):
+        lambda kpath, value: False
+}
```
anusudarsan: I added a couple more properties like plugin, additional_node/config.properties recently. You can take a look at the updated README in presto-yarn

prestoadmin/yarn_slider/server.py
anusudarsan: So in most cases where the application fails (due to some misconfiguration in say, one of the .json files), we do get a success on the create command. But later checking the application status will get a non-zero return code

prestoadmin/yarn_slider/server.py
ebd2: Unless you're thinking of something that I'm missing, if we do that, we need to either:
1. Update config.json with the appname we got from the command line
2. Require that APPNAME be passed in for every command that needs one (stop, destroy)

1 doesn't seem to add much value over the current workflow (user enters app name during the config process), 2 seems cumbersome.

anusudarsan: sounds ok for now. We can revisit how we want to do it when we start to support multiple Presto APPNAMEs (for multiple installations)

prestoadmin/yarn_slider/server.py
ebd2: Oh, I see what you're saying. I think the issue with trying to provide useful status is that it can take several minutes for the app to get into a FAILED state. It seems iffy to expect the user to wait that long unless we can provide some kind of intermediate output, which seems like it'll require digging through the RM REST API and polling the heck out of it.

I don't think waiting for the coordinator to come up is sufficient, because that could work, but then slider could kill the app off later because of worker failures.

prestoadmin/yarn_slider/slider_application_configs.py
ebd2: Sounds like phase 3 unless getting everything else done goes far faster than I expect it to :-)

prestoadmin/yarn_slider/server.py
anusudarsan: True. So the status check could be optional. And if the user doesnt mind waiting then they could do pa slider create --do-status-check. So this flag could do the status check for them after create/start.

prestoadmin/yarn_slider/slider_application_configs.py
ebd2: I'll do that. Thanks for the heads-up.

prestoadmin/yarn_slider/slider_application_configs.py
anusudarsan: :+1:

prestoadmin/yarn_slider/server.py
ebd2: That seems like a reasonable (and highly useful) way to handle the wait issue.

I'm still not clear exactly when we'd declare victory. Do you want to chat tomorrow and see if we can come up with something that both takes a reasonable amount of time and doesn't yield false positives?

anusudarsan: How about using PrestoClient to query if the nodes are up? The same way we do status check in standalone mode

ebd2:
Closing until we decide to commit resources to finishing this work.

PR #100, opened 2015-12-14
Move the installation functionality from test_slider_installation into slider_master

ebd2:
This should reduce the number of circular imports that need to be resolved in
future product tests.

In retrospect, I should have known better :-)


rschlussel:
lgtm

PR #101, opened 2015-12-14
Add test suite for rpm tests

rschlussel:
Presto-admin is not the right place to be testing the presto rpm,
but is currently the only place that can do it.  Still, we don't want
to have to run the tests for the rpm for every change in presto-admin.

Separate tests that are just for the rpm from other product tests and
add a make target for these tests. The rpm tests will not run as part
of make test-all.

Syntax:
make test-rpm

Testing: make lint test-rpm


rschlussel:
@cawallin @ebd2 
Includes tests for https://github.com/Teradata/presto/pull/71

tests/rpm/test_rpm.py
```diff
@@ -0,0 +1,93 @@
...
+from tests.product.standalone.presto_installer import StandalonePrestoInstaller
+
+
+class TestRpm(BaseProductTestCase):
+    def setUp(self):
+
+        super(TestRpm, self).setUp()
+        self.setup_cluster(NoHadoopBareImageProvider(), self.PA_ONLY_CLUSTER)
+
+    @docker_only
```
cawallin: hm, if most of these are going to be docker_only I'd think it would be ok to not have the ability to run rpm tests on a cluster.

tests/rpm/test_rpm.py
```diff
@@ -0,0 +1,93 @@
...
+    def test_install_fails_no_perm_to_local_path(self):
```
cawallin: This is a presto-admin test, not an rpm test

tests/rpm/test_rpm.py
```diff
@@ -0,0 +1,93 @@
...
+        installer.install()
+
+        # starts successfully with java8_home set
+        output = self.run_prestoadmin('server start')
+        self.assertFalse(
+            'Warning: No value found for JAVA8_HOME. Default Java will be '
+            'used.' in output)
+
+    def test_server_starts_no_java8_variable(self):
+        self.setup_cluster(NoHadoopBareImageProvider(),
```
cawallin: If JAVA_HOME is not Java 8, this scenario would fail. At least for the cluster installation, that makes this test too brittle. Maybe one test with and without JAVA_HOME set to a Java 8 path?

cawallin:
looks good % comments

tests/rpm/test_rpm.py
rschlussel: I'm not sure if I tested it where default java was not java 8.  But my understanding was that the server actually starts regardless of what version your java is.  It will then just encounter errors later.

cawallin: I think the launcher script might error out.

tests/rpm/test_rpm.py
rschlussel: k.  I'll get rid of the init script thing and the annotations

tests/rpm/test_rpm.py
rschlussel: removing as not a useful test, as per discussion.

tests/rpm/test_rpm.py
rschlussel: changing test to run on docker only.  Not adding negative test since the rpm doesn't add any checking for this.

ebd2:
Looks good to me!

rschlussel:
merged

PR #103, opened 2015-12-18
Mode switching

ebd2:
Fun times in the world of doing different things for different implementations.
This serves two purposes:
1. This provides a user-friendly way to switch back and forth between managing a YARN-slider managed cluster and a standalone cluster, which is provided with the new 'mode' subcommand.
2. We want the user experience and interface to remain pretty consistent for people managing standalone and YARN-slider managed presto installations. This is done by setting up the imports in __init__.py such that (for example) the commands to start and stop a cluster are exposed through the 'server' subcommand whether you're managing a standalone or YARN-slider managed cluster.

prestoadmin/__init__.py is responsible for importing the right modules and setting up __all__ correctly for main.py to pick it up and load the tasks from the correct modules.

prestoadmin/mode.py is responsible for getting and updating the mode.

Everything else followed from those changes:
* Moving slider install/uninstall into their own module avoided having a bizarre subcommand exception and weird handling for them
* The product tests now need to have a mode.json file installed, so there are now two ModeInstallers.
* Getting the help tests to run was heinous, and getting them to both run in the same test run required some sleuthing to figure out what global state was getting modified (state.env.commands).
* The existing slider product tests had to be updated to use the new subcommands.


ebd2:
Per @cawallin, I'd appreciate a look from her. Also if @rschlussel wouldn't mind taking a peek, I'd appreciate it.

ebd2:
Product tests pass, unit tests pass, lint is clean and docs builds cleanly.

prestoadmin/__init__.py
```diff
@@ -17,30 +17,89 @@
...
+import fabric_patches
+import mode
+
+from prestoadmin.mode import get_mode, for_mode, MODE_STANDALONE, MODE_SLIDER
+from prestoadmin.util.exception import ConfigFileNotFoundError, \
+    ConfigurationError
+
+
+#
+# Subcommands common to all modes. If anybody knows why fabric_patches is in
```
rschlussel: I would guess that it's too avoid warnings about unused imports.  there are better ways to deal with that though.

tests/unit/yarn_slider/test_help.py
```diff
@@ -0,0 +1,80 @@
...
+from mock import patch
+import os
+
+import prestoadmin
+
+from tests.unit.test_main import BaseMainCase
+
+
+#
+# Getting this and TestStandaloneHelp running, and subsequently running
```
rschlussel: Could you make this comment briefer (maybe just paragraphs 2 and 4)?  It's a bit overwhelming.

rschlussel:
I think from a users perspective, it's unexpected that the help test doesn't list all of the things presto-admin does.  Because of the way the task list is loaded for help, there isn't a great way to separately list all tasks for each mode.  But, I think it would be good to add an explicit note above the list of commands that these are the commands for the current mode, but use mode switch to use the other mode, and the commands may be a bit different.

rschlussel:
The rest looks good

ebd2:
I agree about adding note to switch modes to the help, but I think it'd be more confusing to list commands that "don't work" under the current mode. "Don't work" quoted because of the way config is loaded by callback--if you try to run e.g. collect logs, you'll get prompted for the customary topology information, which seems like a terrible thing to have happen for several reasons.

Thanks for taking a look!

Product tests pass, unit tests pass, lint is clean and docs builds cleanly.

tests/unit/yarn_slider/test_help.py
ebd2: How about if I shorten up paragraph 3 to a list of things not to try when refactoring so nobody else goes down a number of dead-end paths like I did yesterday :-(

prestoadmin/__init__.py
ebd2: I suppose the next logical question would be why are we importing it...

tests/unit/yarn_slider/test_help.py
rschlussel: sounds reasonable

prestoadmin/__init__.py
rschlussel: I believe it's imported so that everything in the prestoadmin module has fabric_patches loaded so the function calls use the patched version of all the functions.  It's not really an unused import.  Lint just doesn't know that it's really necessary.  I believe you can configure lint not to care about unused imports in __init__.py, or there is some annotation to say ignore warning for this.

ebd2: Oh, so we can do
from fabric.operations import execute
and we'll actually run the patched version?

rschlussel: yup.  None of the code calls directly to the patched functions.

cawallin: oh yes, I think that Rebecca is remembering correctly.

ebd2:
Updated per Rebecca's feedback. Also, I updated the code in test_main to format the differences between expected command output and actual command output because I've spent far too much time playing Where's Waldo recently.

prestoadmin/__init__.py
```diff
@@ -17,30 +17,89 @@
...
+# the list, I'll make a note for the next person.
+#
+__all__ = ['fabric_patches', 'mode']
+
+cfg_mode = None
+try:
+    cfg_mode = get_mode()
```
cawallin: so there isn't any default? would it make sense to default to rpm installation?
This defnitely needs to be documented and our presto-admin getting started docs need to be updated.

prestoadmin/__init__.py
```diff
@@ -17,30 +17,89 @@
...
+except ConfigFileNotFoundError as e:
+    print >>sys.stderr, \
```
cawallin: Any reason you didn't use a fabric method of printing? That way we can control it with CLI options. (init.py might be too early for fabric things I suppose, though)

prestoadmin/__init__.py
```diff
@@ -17,30 +17,89 @@
...
+                      ('configure_cmds', 'configuration'),
+                      'server',
+                      'connector',
+                      'package',
+                      'collect',
+                      'script',
+                      'plugin']}
+
+
+if cfg_mode is not None:
```
cawallin: A better way to handle this might have been in main.py (though you'd have to mess with how Fabric loads tasks), since having this much logic in init.py is gross, but I'm not going to insist that you change it.

prestoadmin/__init__.py
```diff
@@ -17,30 +17,89 @@
...
+            # The simple case...
+            # import <module> as <subcommand_name>
+            globals()[subcommand_name] = __import__(module, globals())
+        else:
+            # The complicated case:
+            # import foo.bar doesn't actually import foo.bar; it imports foo.
+            # This is why, for example, you can't to the following:
+
+            # >>> import os.path
+            # >>> path.join('foo', 'bar', 'baz', 'zot')
```
cawallin: ah, but you can do from os import path.

prestoadmin/mode.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+    Display the current mode.
+    """
+    mode = None
+    try:
+        mode = get_mode(validate=False)
+    except ConfigFileNotFoundError:
+        pass
+
+    if mode:
```
cawallin: You can merge this if with the above try (print after you get the mode, and in the except abort)

prestoadmin/__init__.py
ebd2: I think you still get the parent module of the leaf module back from __import__ if you pass a fromlist, and since we don't know the name of the module, we'd still have to go fishing in the last parent module's __dict__.

At least that was my interpretation of the docs: https://docs.python.org/2/library/functions.html#__import__

prestoadmin/__init__.py
cawallin: hm, it does make sense to do it here because mode switching is something you want to trick Fabric into doing (and you don't want to have to redefine the mappings every time). Gross.

prestoadmin/mode.py
ebd2: Good use of exceptions, Eric! Fixed.

prestoadmin/__init__.py
ebd2: Yeah, it's not pretty, is it?

prestoadmin/__init__.py
ebd2: I feel like we probably don't want this to be suppressed, but honestly, I didn't even try.

It's worth noting that this gets printed in some of the unit tests too the first time the test_help modules import prestoadmin. Also, I'm highly open to suggestions for getting this to the end of the output. The UX sucks if you run pa with no args: You get this at the top, followed by a bunch of help info. The user shouldn't need to scroll up to see what they need to do to fix stuff.

prestoadmin/__init__.py
ebd2: See my reply about the printing. I'm open to ideas that'll improve the UX.

prestoadmin/__init__.py
cawallin: I doubt you can move this to the end of the output (since it's in init.py rather than in main.py)...

prestoadmin/__init__.py
cawallin: I think defaulting to standalone would be good, that way we don't change how it behaves for current presto-admin users.

ebd2: Will do, and that'll at least put a brand new install into a non user-hostile state.

tests/unit/resources/standalone-help.txt
```diff
@@ -11,6 +11,9 @@ Options:
                         password for use with authentication and/or sudo
 
 
+The commands listed below are for the current mode.
+The current mode can be changed using the 'mode select' command.
```
cawallin: You probably should list the two possible modes here, and maybe also say what mode the user is currently in.

cawallin:
looks good % UX suggestions

tests/unit/resources/standalone-help.txt
ebd2: We're in agreement that the current UX leaves something to be desired, but I'll deal with it next week.

prestoadmin/__init__.py
ebd2: Not easily. main.py calls sys.exit in half a dozen different places :-/

PR #105, opened 2015-12-21
Remove test for install with missing jdk

rschlussel:
There is already a less brittle test for it in the rpm tests

Testing: make lint


cawallin:
looks good

ebd2:
Looks good! We should probably get those rpm tests running on Jenkins against presto-td and presto-fb at some point.

PR #109, opened 2015-12-22
Tests: Fix installer tests passing with a noop

rschlussel:
Set the cluster that the tests should run on.  Remove docker only
requirement since we want to be able to test the installer on other
cluster set ups.

Testing: ran tests

@ebd2


ebd2:
LGTM

rschlussel:
merged

PR #110, opened 2015-12-23
server status command for presto running on yarn-slider

ebd2:
This contains two major pieces:
1. An implementation of the `server status` command for yarn_slider
2. Product tests for the yarn_slider control commands (`server start`, etc)

`server status` is a thin implementation that wraps the built-in Apache slider `status` command. The output is formatted similarly to the output from the existing standalone `server status` command, but information that would be difficult to get is omitted. A more complete implementation is in the works for the future.

The product tests for the control commands are included with this, as they require much of the functionality that server status needs.

Before declaring victory and pushing, I plan to implement a product test for the status command.

Remaining work:
- [x] Product test for status command.

Note that this PR is relative to the server start work, which hasn't yet landed on master. The plan is to hold off on pushing that until this is also ready, as this contains the product tests for server start, then push the two sets of changes at the same time.

@rschlussel, @cawallin, @anusudarsan


prestoadmin/util/slider.py
```diff
@@ -0,0 +1,40 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Utility functions for dealing with slider
+"""
+
+
+def degarbage_json(garbage_output):
+    nest_level = 0
```
rschlussel: 1. could you give this a more descriptive name?  It's not clear that degarbage means remove the non-json part from the beginning and end of the output.
2. You should be able to search for the first occurrence of '{' and last occurrence of '}' e.g. using find() and rfind().  That would be simpler.

tests/unit/util/test_slider.py
```diff
@@ -0,0 +1,68 @@
...
+    def test_simple(self):
+        garbage_in = "Again { the }morning's come"
+        garbage_out = '{ the }'
+        self.assertEqual(garbage_out, degarbage_json(garbage_in))
+
+    def test_no_braces(self):
+        garbage_in = "Again he's on the run"
+        self.assertIsNone(degarbage_json(garbage_in))
+
+    def test_more_left(self):
```
rschlussel: why is it desired to have different behavior for extra braces on the right vs. the left?

rschlussel: I think it should error out for both cases due to malformed json (it's fine if that happens in loads).  It could alternatively be smart in both cases, but that doesn't seem worth the work.

tests/unit/util/test_slider.py
```diff
@@ -0,0 +1,68 @@
...
+        garbage_in = 'Sunbeams shining { { through } his hair'
+        self.assertIsNone(degarbage_json(garbage_in))
+
+    def test_more_right(self):
+        garbage_in = 'Appearing { not { to } have } a } care'
+        garbage_out = '{ not { to } have }'
+        self.assertEqual(garbage_out, degarbage_json(garbage_in))
+
+    def test_second_json(self):
```
rschlussel: seems arbitrary to choose the first over the second.  I think it's fine to error out in this case too.  But if not, maybe put a comment in the method about expected behavior.

prestoadmin/yarn_slider/server.py
```diff
@@ -248,3 +250,79 @@ def mk_data_dir():
...
+def get_instances(slider_status, rolename):
+    return slider_status['instances'][rolename]
+
+
+def identity(x):
+    return x
+
+
+def get_coordinator(slider_status, filtr=identity):
+    coordinators = slider_status['status']['live'][ROLENAME_COORDINATOR]
```
rschlussel: If there's something wrong on the cluster is there status for anything other than 'live' that appears?  is it worth showing?

prestoadmin/yarn_slider/server.py
```diff
@@ -248,3 +250,79 @@ def mk_data_dir():
...
+    conf = env.conf
+    slider_status = get_slider_status(conf)
+    coordinators = None
+
+    try:
+        coordinators = get_coordinator(slider_status)
+    except KeyError:
+        abort(
+            'No coordinator found. This may be because presto is still '
+            'starting. Check the application status')
```
rschlussel: tell people how to check the application status

rschlussel: also why abort if there's no coordinator.  Does is it prevent you from knowing anything useful about the workers too?

prestoadmin/util/slider.py
ebd2: 1. Guess who was a Grumpy Gus when he was writing this :-) I'll change it. Slider doesn't separate the JSON from its messages into stdout and stderr.
2. That would be simpler, but it doesn't guarantee that you end up with a balanced bunch of braces. Mind you, this isn't doing anything else to ensure that it returns well-formed JSON...

prestoadmin/yarn_slider/server.py
ebd2: Unclear; this falls into the usual problem of getting something into a messed-up state long enough to determine the expected behavior. I'll dig into the Slider code and see if I can pull out something useful.

prestoadmin/yarn_slider/server.py
ebd2: I've only seen this happen in the short period between starting the app instance and it coming up (which is why there are a bunch of places where we wait for presto to come up and go down in the new product tests).

How does the following sound:
1. Add information about the workers, if available.
2. Change the error to tell the user to try wait a minute and try again or check the status via the ResourceManager and give the URL.
3. Possibly add another command to dump the raw slider JSON for debugging?

prestoadmin/util/slider.py
ebd2: The other possibility, by the way, is to use Slider's -out option which can be used to get Slider to output clean JSON to a file. That option circumvents the need to try to pluck some JSON out of a string, but comes with other problems.

Notably, we need to be able to write a file to a sufficiently unique path that nothing blows up if multiple people are getting the status at the same time. Then we need to clean it up, which is simple enough if everything is going well, but risks cluttering up /tmp if pa somehow crashes between having slider write the file and deleting it. It goes without saying that the only thing we should be doing in that window is trying to grab the file via fabric, but Murphy's law and all that.

prestoadmin/yarn_slider/server.py
ebd2: After getting the slider source and doing some digging, the answer appears to be no, at least not currently. There's a key for "failed" as well, but it isn't used in the codebase.

prestoadmin/util/slider.py
rschlussel: I like that option better.  python will give you uniquely named temporary files, so that shouldn't be a problem.  And I don't think this would be called frequently enough to really worry about filling up the tmp directory if it crashes.  It would just be one stray file left behind in rare circumstances, and in a tmp directory.  
But I believe we could also instead do something like 'bin/slider status -out > cat' to print the output to stdout instead of saving it to a file.

prestoadmin/yarn_slider/server.py
rschlussel: 1 and 2 sound good.  3 sounds useful but not necessary.

prestoadmin/util/slider.py
ebd2: After talking about this offline, we think we have an approach that doesn't require plucking the JSON out of a bunch of other output or cleaning up temp files on the slider master:
We're going to use -out, but we'll use /dev/stdout as the output file.
Things that don't work:
* `--out -`. slider doesn't follow the convention that a - for the output path means stdout
* process substitution `--out >(cat)`. Slider errors out with the following: `main.ServiceLauncher - java.io.FileNotFoundException: /dev/fd/63 (Permission denied)`

/dev/stdout seems to be sufficiently portable for our use:
http://unix.stackexchange.com/questions/36403/portability-of-dev-stdout

anusudarsan: I like the idea of using --out too. presto-yarn product tests uses it too. I think --out option works only for outputting to a file, like --out <filepath>

prestoadmin/yarn_slider/server.py
```diff
@@ -248,3 +250,79 @@ def mk_data_dir():
...
+
+
+def get_coordinator(slider_status, filtr=identity):
+    coordinators = slider_status['status']['live'][ROLENAME_COORDINATOR]
+    return [filtr(v) for v in coordinators.values()]
+
+
+def get_workers(slider_status, filtr=identity):
+    try:
+        workers = slider_status['status']['live'][ROLENAME_WORKER]
```
anusudarsan: maybe extract the slider_status [] ['live'][component_name] to a method x and call x(ROLENAME_WORKER) and x(ROLENAME_COORDINATOR)

prestoadmin/util/slider.py
rschlussel: the idea here was to use /dev/stdout as the file name so you don't have to worry about cleaning up temporary files.

prestoadmin/yarn_slider/server.py
anusudarsan: slider wont deploy any workers till coordinator is up. So I doubt if you will get any useful info for case 1.

prestoadmin/util/slider.py
anusudarsan: ah. I missed the last section in the comment about /dev/stdout. that shud work.

ebd2:
Closing until we decide to commit resources to finishing this work.

PR #112, opened 2015-12-23
Collect even if presto-admin not on presto node

rschlussel:
Fix a bug where collect system_info and collect query_info always
queried local host, and so didn't work when presto-admin was not on a
presto node.

Testing: make lint test, product tests in test_collect

@ebd2 tests are still running


ebd2:
Looks good to me pending a successful test run. I've already created the release branch, so this'll have to be back-ported to it. Sorry about that :-(

PR #116, opened 2016-01-15
Changes to the plugin.py file to include the plugin remove_jar feature

nanaessuman:


prestoadmin/plugin.py
```diff
@@ -47,3 +46,15 @@ def add_jar(local_path, plugin_name, plugin_dir=REMOTE_PLUGIN_DIR):
     """
     _LOGGER.info('deploying jars on %s' % env.host)
     write(local_path, os.path.join(plugin_dir, plugin_name))
+
+def remove_jar(local_path, plugin_name):
```
rschlussel: For remove jar, there is no local path.  The command should take the plugin name to remove and optionally the remote path to the plugin directory to allow for non-default configurations.

def remove_jar(plugin_name, plugin_dir=REMOTE_PLUGIN_DIR)

prestoadmin/plugin.py
```diff
@@ -47,3 +46,15 @@ def add_jar(local_path, plugin_name, plugin_dir=REMOTE_PLUGIN_DIR):
...
+    """
+    Remove jar for the specified plugin to the plugin directory.
+
+    Parameters:
+        local_path - Local path to the jar to be removed
+        plugin_name - Name of the plugin subdirectory to deploy jars to
+
+    """
+    _LOGGER.info('removing jar on %s' % env.host)
+    os.remove(os.path.join(local_path, plugin_name))
```
rschlussel: This will only remove the jar from the local system the presto-admin is running on.  To remove it from all of the presto nodes you should use a fabric command something like.

path = os.path.join(plugin_dir, plugin_name)
sudo('rm %s' % path)

rschlussel:
Thanks for the patch! We're excited to have a new contributor!
A couple notes on testing:
You should  run 'make lint' to ensure that the code adheres to our style. 
It's best to test with at least one presto node on a different node from presto-admin.  Usually we write product tests to be able to run this automatically, but it's a bit complicated to set up, so it's fine if you just verify manually.

kokosing:
@nanaessuman Could you please address the comments from @rschlussel ? So we could merge this PR.

PR #117, opened 2016-01-18
Use .* to check version string.

rschlussel:
Presto changed the way it displays the server version.  We don't care
what the version string is, so just match against .*


tests/product/base_product_case.py
```diff
@@ -36,7 +36,7 @@
 from tests.product.topology_installer import TopologyInstaller
 from tests.product.yarn_slider.slider_installer import SliderInstaller
 
-PRESTO_VERSION = r'presto-main:.*'
+PRESTO_VERSION = r'.*'
```
petroav: Maybe change it to `r'.+'` so that at least one character is required. Otherwise, the current regex will match an empty string.

petroav:
Other than my comment, lgtm.

PR #119, opened 2016-01-19
Update Makefile help and add clean-docs target

petroav:


cawallin:
looks good

Makefile
```diff
@@ -51,6 +56,11 @@ clean-test:
 	-for image in $$(docker images | awk '/teradatalabs\/pa_test/ {print $$1}'); do docker rmi -f $$image ; done
 	@echo "\n\tYou can kill running containers that caused errors removing images by running \`make clean-test-containers'\n"
 
+clean-docs:
```
ebd2: clean should probably also depend on clean-docs
clean-docs should be .PHONY. See line 1. Actually, all of the clean targets should be .PHONY. Could you please add them while you're fixing this?

petroav: Done.

PR #122, opened 2016-01-29
Fetch more log lines for test_status failures.

ebd2:
The presto-admin status command logs all the things. Get more lines on
failures so we have a better picture of what's happening when we execute
queries.


cawallin:
looks good

PR #124, opened 2016-02-08
dist: Pin setuptools at 19.7

cawallin:
Setuptools 20.0 does not work on CentOS.


cawallin:
Sandbox build in progress

cawallin:
oops, I had another change along with this one; this one alone doesn't work.

PR #126, opened 2016-02-12
Fixes for intermittent os error

rschlussel:
The first commit just refactors tests so that they do one thing.  A lot of the instability was happening during clean up from one test to prepare for another, which highlighted some improvements that could be made in the tests

The second commit adds a retry to assert_stopped.  The service commands are backgrounded for shell/fabric/centos reasons (https://github.com/prestodb/presto-admin/commit/35823e083ba894a5482685d53d99ad290833d10e), meaning the tasks return before they are necessarily done.  Since we can't get a return code to know when they are done, we add a wait.

### DIFF
--- a/tests/product/test_control.py
+++ b/tests/product/test_control.py
@@ -58,35 +58,29 @@ class TestControl(BaseProductTestCase):
         self.assertEqualIgnoringOrder(presto_not_installed,
                                       '\n'.join(start_output))
 
-    def test_server_start_various_states(self):
+    def test_server_start_one_host_started(self):
         self.setup_cluster(NoHadoopBareImageProvider(),
                            self.STANDALONE_PRESTO_CLUSTER)
-
-        # Coordinator started, workers not; then server start
-        process_per_host = \
-            self.assert_start_with_one_host_started(
-                self.cluster.internal_master)
-
-        # Worker started, coord and other workers not; then server start
-        self.run_prestoadmin('server stop').splitlines()
-        self.assert_stopped(process_per_host)
         self.assert_start_with_one_host_started(
             self.cluster.internal_slaves[0])
 
-        # All started; then server start
+    def test_server_start_all_started(self):
+        self.setup_cluster(NoHadoopBareImageProvider(),
+                           self.STANDALONE_PRESTO_CLUSTER)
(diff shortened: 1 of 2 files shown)


tests/product/test_control.py
```diff
@@ -58,35 +58,35 @@ def assert_service_fails_without_presto(self, service):
...
-    def test_server_start_various_states(self):
+    def test_server_start_coordinator_started(self):
         self.setup_cluster(NoHadoopBareImageProvider(),
                            self.STANDALONE_PRESTO_CLUSTER)
 
-        # Coordinator started, workers not; then server start
-        process_per_host = \
-            self.assert_start_with_one_host_started(
-                self.cluster.internal_master)
+        self.assert_start_with_one_host_started(self.cluster.internal_master)
```
cawallin: the point of this test was to make sure that if you stop with one host already started, it works; you should add a stop here

tests/product/test_control.py
```diff
@@ -58,35 +58,35 @@ def assert_service_fails_without_presto(self, service):
...
 
-        # Worker started, coord and other workers not; then server start
-        self.run_prestoadmin('server stop').splitlines()
-        self.assert_stopped(process_per_host)
+    def test_server_start_worker_started(self):
+        self.setup_cluster(NoHadoopBareImageProvider(),
+                           self.STANDALONE_PRESTO_CLUSTER)
         self.assert_start_with_one_host_started(
```
cawallin: The point for this one was that start works if one host is already started; you should add a start all after this

tests/product/test_control.py
cawallin: j/k, I think this wasn't actually doing anything helpful. If we want to separately test "start coord -> start all" and "start worker -> start all" we could keep this, otherwise we don't need it

cawallin:
a few comments, otherwise looks good!

tests/product/test_control.py
rschlussel: assert_start_with_one_host_started first starts that one host and then starts all of them.  Agree that we probably only need coordinator or workers and not both.  I'll remove one of the tests and change to test_server_start_one_host_started().

PR #127, opened 2016-02-17
Tests: add retry for status when connectors absent

rschlussel:
Testing: test_status

@cawallin @ebd2


ebd2:
Looks good to me. Do we want to drop the number of log lines we fetch on failure or should we wait and see things stabilize first?

cawallin:
looks good

rschlussel:
Maybe drop it.  I didn't see anything useful in it (except noting an absence of error messages).

PR #128, opened 2016-02-17
Update readme with new dist behavior

rschlussel:
@cawallin @ebd2 
I didn't put anything about pycrypto since that's still a bit of a mess.  But I updated the readme to be accurate


cawallin:
looks good, thanks!

PR #129, opened 2016-02-23
Add more debugging info for when config is missing

rschlussel:
The first commit is unrelated, but super tiny.

The second commit changing the arguments to assertLazyMessage() was a relic of an earlier way that I had added debugging that required arguments beyond actual and expected.  It's no longer needed for my patches, but it seemed independently valuable since it makes the function more flexible, and I already tested it, so I left it in.  If you prefer it the old way though, I'm happy to revert.

@ebd2

### DIFF
--- a/tests/base_test_case.py
+++ b/tests/base_test_case.py
@@ -115,9 +115,9 @@ class BaseTestCase(unittest.TestCase):
             self.fail("Expected exception " + str(expected_exception) +
                       " not raised" + msg)
 
-    def assertLazyMessage(self, assert_function, actual, expected, msg_func):
+    def assertLazyMessage(self, msg_func, assert_function, *args, **kwargs):
         try:
-            assert_function(actual, expected)
+            assert_function(*args, **kwargs)
         except AssertionError:
             self.fail(msg=msg_func())
 
--- a/prestoadmin/workers.py
+++ b/prestoadmin/workers.py
@@ -60,7 +60,6 @@ class Worker(Node):
         return constants.WORKERS_DIR
 
     def default_config(self, filename):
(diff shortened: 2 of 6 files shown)


cawallin:
looks good

tests/base_test_case.py
```diff
@@ -115,9 +115,9 @@ def assertRaisesRegexp(self, expected_exception, expected_regexp,
             self.fail("Expected exception " + str(expected_exception) +
                       " not raised" + msg)
 
-    def assertLazyMessage(self, assert_function, actual, expected, msg_func):
+    def assertLazyMessage(self, msg_func, assert_function, *args, **kwargs):
```
ebd2: We could probably pass *args and **kwargs into msg_func and lose all the silly lambdas I created. I apparently did not have my python hat on when I wrote this originally. If you want to push without this change and file a ticket, I can worry about it later.

ebd2: I guess that couples the msg_func more tightly with the assertion though.
"All problems in computer science can be solved by another level of indirection"/"...except for the problem of too many layers of indirection." -- Wheeler/Henney

ebd2:
LGTM modulo matters of taste. I'm fine with you pushing it as is and punting the matter of taste into the future when we have more examples to generalize from.

tests/base_test_case.py
rschlussel: I'm going to push it as is now.  The arguments to msg_func currently aren't all the same as those to the assert_function, so there isn't an easy 2 minute change, and I want to get the extra debugging in now in hopes of capturing something sooner.   But I agree that some sort of future clean up would be good.

rschlussel:
merged

PR #131, opened 2016-02-23
Make container build fail if ssh key fails to download

cawallin:


cawallin:
Running sandbox build now

rschlussel:
Why doesn't it fail when the download fails?

cawallin:
hm, I guess it should fail if the download doesn't work. removing the if statement shouldn't hurt though.

rschlussel:
yup.  This looks good.

PR #132, opened 2016-02-23
Change tests to print node.properties when no logs

rschlussel:
The log location is set in node.properties, not log.properties.  Fix
debug info

I also realized that it will go to /var/log/presto by default (for the rpm), so I think it's probably not related to the missing config issue.  but i left this logging in as something that could be maybe possibly helpful.  

@ebd2


ebd2:
LGTM

rschlussel:
merged, thanks!

PR #133, opened 2016-02-24
Change default jvm.config to be 16G rather than 2G

cawallin:
Our recommendations for Presto memory are to use 16GB, so we need to
change presto-admin to do that.


cawallin:
FYI @KBP-TDC

ebd2:
LGTM.

docs/presto-admin-commands.rst
```diff
@@ -113,7 +113,7 @@ configurations:
 *jvm.config* ::
 
     -server
-    -Xmx2G
+    -Xmx16G
```
akshatnair: Should we correspondingly also increase the values for query.max_memory_per_node?

cawallin: ah yes. probably.

PR #134, opened 2016-02-24
Raise error for server and connector failures

rschlussel:
Server start and connector add/remove had been giving a warning instead of
throwing an error when a task failed.  Now they throw an error and exit
with a non-zero exit code.

Testing: make lint test-all

@ebd2 @cawallin

### DIFF
--- a/prestoadmin/connector.py
+++ b/prestoadmin/connector.py
@@ -18,7 +18,7 @@ Module for presto connector configurations
 import logging
 import errno
 
-from fabric.api import task, env, abort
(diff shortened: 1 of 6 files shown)


cawallin:
If one of the tasks fails, does it stop all the rest of the tasks from running?

rschlussel:
You mean on the other nodes?  Those will still complete.

cawallin:
ok, looks good

ebd2:
LGTM

rschlussel:
merged

PR #140, opened 2016-03-07
test_connectors: Update expected results

cawallin:
Since the Fabric calls to different slaves are asynchronous, "Aborting"
can print out anywhere (e.g. we could have "Fatal error... Fatal error...
Aborting. Aborting). We switch to use the ignoringOrder method of comparison
instead of regex comparision.


rschlussel:
lgtm

PR #144, opened 2016-03-09
Fix lint and requirements

cawallin:


cawallin:
@kokosing @rschlussel

setup.py
```diff
@@ -16,7 +16,7 @@
 import os
 
 # This is necessary for nose to handle multiprocessing correctly
-from multiprocessing import util
+from multiprocessing import util  # noqa
```
rschlussel: what is the flake8 error that would get raised here?

cawallin: unused import, but I wasn't sure about removing it b/c of the comment

rschlussel:
looks good.

kokosing:
looks good

Maybe you could make line length limit longer just for convenience. Now you had to split so many lines, with longer line length limit you could avoid that.

cawallin:
We were hesitant to change the line length limit, since that's the standard PEP-8 limit. Might as well keep it for now.

PR #146, opened 2016-03-14
fixup! Wait for presto user after rpm installation

cawallin:
Fix the unit test so that it doesn't hang forever in checking for the Presto user.

### DIFF
--- a/tests/unit/test_server.py
+++ b/tests/unit/test_server.py
@@ -53,23 +53,24 @@ class TestInstall(BaseUnitCase):
         mock_execute.assert_called_with(server.deploy_install_configure,
                                         local_path, hosts=get_host_list())
 
+    @patch('prestoadmin.server.sudo')
     @patch('prestoadmin.server.package.deploy_install')
     @patch('prestoadmin.server.update_configs')
-    def test_deploy_install_configure(self, mock_update, mock_install):
+    def test_deploy_install_configure(self, mock_update, mock_install,
+                                      mock_sudo):
         local_path = "/any/path/rpm"
+        mock_sudo.side_effect = self.mock_fail_then_succeed()
+
         server.deploy_install_configure(local_path)
         mock_install.assert_called_with(local_path)
-        mock_update.assert_called_with()
+        self.assertTrue(mock_update.called)
+        mock_sudo.assert_called_with('getent passwd presto')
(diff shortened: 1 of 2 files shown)


cawallin:
FYI @kokosing

rschlussel:
lgtm

tests/unit/test_server.py
```diff
@@ -53,23 +53,23 @@ def test_install_server(self, mock_execute, mock_check_rpm):
...
 
+    @patch('prestoadmin.server.sudo')
     @patch('prestoadmin.server.package.deploy_install')
     @patch('prestoadmin.server.update_configs')
-    def test_deploy_install_configure(self, mock_update, mock_install):
+    def test_deploy_install_configure(self, mock_update, mock_install,
+                                      mock_sudo):
         local_path = "/any/path/rpm"
+        mock_sudo.side_effect = self.mock_failed_then_succeeded()
         server.deploy_install_configure(local_path)
```
petroav: Add a newline after this to separate the test from the assertion statements.

tests/unit/test_server.py
```diff
@@ -53,23 +53,23 @@ def test_install_server(self, mock_execute, mock_check_rpm):
...
         mock_install.assert_called_with(local_path)
         mock_update.assert_called_with()
```
petroav: If we're not checking the arguments used to call the mock maybe we can substitute this call with something simpler?

tests/unit/test_server.py
```diff
@@ -454,3 +454,10 @@ def test_multiple_version_rpms(self, mock_run):
             call('rpm -q --qf \"%{VERSION}\\n\" presto-server-rpm'),
         ])
         self.assertEqual(expected, '')
+
+    def mock_failed_then_succeeded(self):
```
petroav: I prefer `mock_fail_then_succeed`. Your call.

petroav:
looks good

tests/unit/test_server.py
cawallin: Done

tests/unit/test_server.py
cawallin: Sure, changed it to "succeeded" to match the attribute, but that's fine.

PR #147, opened 2016-03-15
install : create dirs useful for prestoadmin

akshatnair:
We need dirs like connectors, coodinator and workers
created in the location /etc/opt/prestoadmin. Pre-creating
them will give a better user experience for configuring Presto
using prestoadmin.

Testing: added a new test

### DIFF
--- a/packaging/install-prestoadmin.template
+++ b/packaging/install-prestoadmin.template
@@ -61,3 +61,12 @@ mkdir -p $LOG_DIR
 
 CONF_DIR=/etc/opt/prestoadmin
 mkdir -p $CONF_DIR
+
+CONNECTOR_DIR=$CONF_DIR/connectors
+mkdir -p $CONNECTOR_DIR
+
+COORDINATOR_DIR=$CONF_DIR/coordinator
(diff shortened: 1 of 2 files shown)


rschlussel:
lgtm

packaging/install-prestoadmin.template
```diff
@@ -61,3 +61,12 @@ mkdir -p $LOG_DIR
 
 CONF_DIR=/etc/opt/prestoadmin
 mkdir -p $CONF_DIR
+
+CONNECTOR_DIR=$CONF_DIR/connectors
+mkdir -p $CONNECTOR_DIR
```
cawallin: Who owns these directories? root?
It's ok whatever it is, but @rschlussel should keep it in mind for her work to remove the requirement to run as sudo.

cawallin:
lgtm

packaging/install-prestoadmin.template
akshatnair: yeah it is the same as /etc/opt/prestoadmin

akshatnair:
merged

PR #148, opened 2016-03-15
Add tests for -H and -x input arguments

amrutagokhale:
I added an initial commit for positive tests for 'connector add -H' and 'connector remove -H'. I am nowhere near adding all the tests. But it would be nice if I get some feedback so as to know if I am on the right track.

I will add more commits as I write more tests.

### DIFF
--- a/tests/product/test_collect.py
+++ b/tests/product/test_collect.py
@@ -63,14 +63,6 @@ class TestCollect(BaseProductTestCase):
         admin_log = path.join(downloaded_logs_location, PRESTOADMIN_LOG_NAME)
         self.assert_path_exists(self.cluster.master, admin_log)
 
-    def test_collect_logs_dash_h_invalid_host(self):
-        self.setup_cluster(NoHadoopBareImageProvider(),
-                           self.STANDALONE_PRESTO_CLUSTER)
-        expected = 'Hosts defined in --hosts/-H must be present in /etc/opt/prestoadmin/config.json'
-
-        self.assertRaisesRegexp(OSError, expected, self.run_prestoadmin,
-                                'collect logs -H invalidHost' )
-
     def log_msg(self, actual, expected):
         msg = '%s != %s' % actual, expected
         return msg
@@ -123,23 +115,23 @@ class TestCollect(BaseProductTestCase):
         self.setup_cluster(NoHadoopBareImageProvider(),
                            self.STANDALONE_PRESTO_CLUSTER)
         self.run_prestoadmin('server start')
-        actual = self.run_prestoadmin('collect system_info -H %(master)s,%(slave1)s')
-        self.test_basic_system_info(actual, self.cluster.internal_master, [self.cluster.master, self.cluster.slaves[0]])
+        actual = self.run_prestoadmin('collect system_info '
+                                      '-H %(master)s,%(slave1)s')
+        self.test_basic_system_info(actual,
+                                    self.cluster.internal_master,
+                                    [self.cluster.master,
+                                     self.cluster.slaves[0]])
 
     def test_collect_system_info_dash_x_two_workers(self):
         self.setup_cluster(NoHadoopBareImageProvider(),
                            self.STANDALONE_PRESTO_CLUSTER)
         self.run_prestoadmin('server start')
-        actual = self.run_prestoadmin('collect system_info -x %(slave2)s,%(slave3)s')
-        self.test_basic_system_info(actual, self.cluster.internal_master, [self.cluster.master, self.cluster.slaves[0]])
-
-    def test_collect_system_info_dash_h_invalid_host(self):
-        self.setup_cluster(NoHadoopBareImageProvider(),
-                           self.STANDALONE_PRESTO_CLUSTER)
-        expected = 'Hosts defined in --hosts/-H must be present in /etc/opt/prestoadmin/config.json'
-
-        self.assertRaisesRegexp(OSError, expected, self.run_prestoadmin,
-                                'collect system_info -H invalidHost' )
+        actual = self.run_prestoadmin('collect system_info '
+                                      '-x %(slave2)s,%(slave3)s')
+        self.test_basic_system_info(actual,
+                                    self.cluster.internal_master,
+                                    [self.cluster.master,
+                                     self.cluster.slaves[0]])
 
     def test_system_info_pa_separate_node(self):
         installer = StandalonePrestoInstaller(self)
@@ -148,7 +140,10 @@ class TestCollect(BaseProductTestCase):
                     "workers": ["slave2", "slave3"]}
         self.upload_topology(topology=topology)
         installer.install(coordinator='slave1')
+        self.run_prestoadmin('server start')
+        actual = self.run_prestoadmin('collect system_info')
         self.test_basic_system_info(
+            actual,
             coordinator=self.cluster.internal_slaves[0],
             hosts=self.cluster.slaves)
 
--- a/tests/product/test_collect.py
+++ b/tests/product/test_collect.py
@@ -63,6 +63,14 @@ class TestCollect(BaseProductTestCase):
         admin_log = path.join(downloaded_logs_location, PRESTOADMIN_LOG_NAME)
         self.assert_path_exists(self.cluster.master, admin_log)
 
+    def test_collect_logs_dash_h_invalid_host(self):
+        self.setup_cluster(NoHadoopBareImageProvider(),
+                           self.STANDALONE_PRESTO_CLUSTER)
+        expected = 'Hosts defined in --hosts/-H must be present in /etc/opt/prestoadmin/config.json'
+
+        self.assertRaisesRegexp(OSError, expected, self.run_prestoadmin,
+                                'collect logs -H invalidHost' )
+
     def log_msg(self, actual, expected):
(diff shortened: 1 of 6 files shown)


tests/product/test_connectors.py
```diff
@@ -61,6 +61,96 @@ def test_basic_connector_add_remove(self):
...
+                           self.STANDALONE_PRESTO_CLUSTER)
+        self.run_prestoadmin('server start')
+        for host in self.cluster.all_hosts():
+            self.assert_has_default_connector(host)
+
+        self._assert_connectors_loaded([['system'], ['tpch']])
+
+        self.run_prestoadmin('connector remove tpch -H %(master)s')
+        self.run_prestoadmin('server restart')
+        self.assert_path_removed(self.cluster.master,
```
rschlussel: I see what you mean about -H not making sense.  It's really weird behavior to delete the presto-admin connector config file when you are only removing the connector on one node.  But I still stand by what I said before.

tests/product/test_connectors.py
```diff
@@ -61,6 +61,96 @@ def test_basic_connector_add_remove(self):
...
+            'connector.name=tpch',
+            os.path.join(constants.CONNECTORS_DIR, 'tpch.properties'),
+            self.cluster.master
+        )
+        self.run_prestoadmin('connector add tpch -H %(slave1)s')
+        self.run_prestoadmin('server restart')
+        self.assert_has_connector(self.cluster.slaves[0], 'tpch')
+        self._assert_connectors_loaded([['system'], ['tpch']])
+
+    def test_basic_connector_add_remove_workers_using_dash_h(self):
```
rschlussel: I would make this name more distinct from the one above that tests one worker.  The extra s is easy to miss.  Also I think the word 'basic' isn't particularly informative.  I know lots of tests are named that way, but for brevity's sake I'd get rid of it in these tests.  It's anyway not basic anymore once you are using -H.

tests/product/test_connectors.py
amrutagokhale: Yes, I believe that in the long run, we should think about whether -H makes sense for each of the commands, and if it doesn't, we should error out. I will add tests for now anyways.

rschlussel:
these look good.  definitely the right track!

tests/product/test_connectors.py
```diff
@@ -61,6 +61,96 @@ def test_basic_connector_add_remove(self):
...
+            'connector.name=tpch',
+            os.path.join(constants.CONNECTORS_DIR, 'tpch.properties'),
+            self.cluster.master
+        )
+        self.run_prestoadmin('connector add tpch -H %(slave1)s')
+        self.run_prestoadmin('server restart')
+        self.assert_has_default_connector(self.cluster.slaves[0])
+        self._assert_connectors_loaded([['system'], ['tpch']])
+
+    def test_basic_connector_add_remove_workers_using_dash_h(self):
```
rschlussel: I would make this name more distinct from the one above that tests one worker. The extra s is easy to miss. Also I think the word 'basic' isn't particularly informative. I know lots of tests are named that way, but for brevity's sake I'd get rid of it in these tests. It's anyway not basic anymore once you are using -H.

tests/product/test_connectors.py
amrutagokhale: done

tests/product/test_connectors.py
```diff
@@ -61,6 +61,96 @@ def test_basic_connector_add_remove(self):
...
+            'connector.name=tpch',
+            os.path.join(constants.CONNECTORS_DIR, 'tpch.properties'),
+            self.cluster.master
+        )
+        self.run_prestoadmin('connector add tpch -H %(slave1)s')
+        self.run_prestoadmin('server restart')
+        self.assert_has_default_connector(self.cluster.slaves[0])
+        self._assert_connectors_loaded([['system'], ['tpch']])
+
+    def test_connector_add_remove_two_workers_using_dash_h(self):
```
kokosing: I don't know why you have 3 tests of the same functionality. Does it matter what node you add or remove? I think the last tests is enough. I just would mix slave with master, and one test could be enough. Less tests, faster it is executed, less code needs to be maintained.

What do you think?

ebd2: Agree wholeheartedly. The product tests are a very expensive way to find bugs. This tests the behavior of the command line parsing more than the actual product tests.

amrutagokhale: Good point. I agree that the last test should be sufficient. The latest commit retains only the last test.

kokosing:
looks good

amrutagokhale:
@kokosing @ebd2 Added three more commits. Can you review?

tests/product/test_collect.py
```diff
@@ -64,6 +64,14 @@ def test_collect_logs_basic(self):
         admin_log = path.join(downloaded_logs_location, PRESTOADMIN_LOG_NAME)
         self.assert_path_exists(self.cluster.master, admin_log)
 
+    def test_collect_logs_dash_h_invalid_host(self):
+        self.setup_cluster(NoHadoopBareImageProvider(),
+                           self.STANDALONE_PRESTO_CLUSTER)
+        expected = 'Hosts defined in --hosts/-H must be present in /etc/opt/prestoadmin/config.json'
```
ebd2: This looks too long. Have you run `make lint`?

tests/product/test_collect.py
```diff
@@ -131,6 +139,28 @@ def test_basic_system_info(self, coordinator=None, hosts=None):
...
+        self.test_basic_system_info(actual, self.cluster.internal_master, [self.cluster.master, self.cluster.slaves[0]])
+
+    def test_collect_system_info_dash_x_two_workers(self):
+        self.setup_cluster(NoHadoopBareImageProvider(),
+                           self.STANDALONE_PRESTO_CLUSTER)
+        self.run_prestoadmin('server start')
+        actual = self.run_prestoadmin('collect system_info -x %(slave2)s,%(slave3)s')
+        self.test_basic_system_info(actual, self.cluster.internal_master, [self.cluster.master, self.cluster.slaves[0]])
+
+    def test_collect_system_info_dash_h_invalid_host(self):
```
ebd2: Validating the command line arguments should be tested in a unit test; this shouldn't vary from command to command, and it *definitely* shouldn't rely on having a cluster up.

tests/product/test_configuration.py
```diff
@@ -107,6 +115,70 @@ def test_configuration_deploy_show(self):
         self.assert_node_config(self.cluster.master,
                                 self.default_node_properties_)
 
+    def test_configuration_deploy_using_dash_h_coord_worker(self):
+        self.upload_topology()
+
+        self.deploy_and_assert_default_config()
+
+        dummy_prop1, dummy_prop2 = self.write_dummy_config_file()
```
ebd2: Ooh, I like this! Nice touch!

ebd2:
Apart from getting rid of the invalid host cases and writing a single unit test, this looks good.

tests/product/test_configuration.py
```diff
@@ -58,15 +55,26 @@ def test_configuration_deploy_show(self):
 
         self.assertEqualIgnoringOrder(output, expected)
 
+    def write_dummy_config_file(self):
```
kokosing: make this private by prefixing with `__`

kokosing:
lgtm

tests/product/test_collect.py
kokosing: +1

tests/product/test_connectors.py
```diff
@@ -90,6 +90,22 @@ def test_connector_add_remove_coord_worker_using_dash_h(self):
         self.assert_has_default_connector(self.cluster.master)
         self.assert_has_default_connector(self.cluster.slaves[1])
 
+    def test_connector_add_invalid_host_using_dash_h(self):
```
kokosing: these two tests looks like they could be written at unit test level as they are just parsing arguments

tests/product/test_configuration.py
amrutagokhale: done

tests/product/test_collect.py
amrutagokhale: That's a good suggestion. I am now adding a single unit test for one sample command `server uninstall -H <invalid host name>` and I am going to remove all the product tests for `-H <invalid host name>`. There is no particular reason for choosing `server uninstall` command. I had to choose one command to which I can provide `-H` option. `server uninstall` already had an existing unit test that checks for invalid host name using --hosts: `server uninstall --hosts <invalid host name>`. So I added the test for the shorter `-H` option for the same command.

tests/product/test_collect.py
amrutagokhale: This is being removed and replaced by a single unit test.

tests/product/test_connectors.py
amrutagokhale: Yes, I have removed these as well. Now, there exists a single unit test that checks for `-H <invalid host name>`

amrutagokhale:
updated

kokosing:
thanks for applying comments, LGTM :)

kokosing:
Fix travis build, it looks like `lint` found some issues.

amrutagokhale:
Merged https://github.com/prestodb/presto-admin/pull/148

PR #153, opened 2016-03-16
Remove dummy-rpm.rpm usage from product tests

kokosing:
Remove dummy-rpm.rpm usage from product tests

Mocking rpm file in product tests does not much black-box testing.


losipiuk:
lgtm - but please fix the commit message.

ebd2:
I don't love the amount of time we're going to be adding to the test runs, but I'm not running these locally at the moment, and we can probably address the by taking another look at the tests instead of using a dummy RPM.

LGTM.

cawallin:
lgtm

ebd2:
Updated commit message and merged.

PR #155, opened 2016-03-16
Add dummy RPM back in for upgrade product tests.

ebd2:
It turns out we actually need the dummy RPM for the upgrade product tests. The
upgrade tests don't depend on any of the random RPM functionality that changes
periodically, and we need *some* RPM to do an upgrade. Since downloading
another 400MB from the internet for one test is onerous, we use the dummy RPM.


tests/product/test_server_upgrade.py
```diff
@@ -63,7 +77,7 @@ def test_upgrade(self):
 
         self.run_prestoadmin('configuration deploy')
         for container in self.cluster.all_hosts():
-            self.installer.assert_installed(self, container)
+            self.real_installer.assert_installed(self, container)
```
cawallin: presumably real_installer would no longer be installed?

cawallin: oh, sorry, this is not a part of the previous method. that's fine.

cawallin:
looks good

PR #157, opened 2016-03-17
Add precommit target

ebd2:
If we only had a dollar for every time somebody (me included) pushed with a
lint or sphinx error. make precommit runs all the stuff that's quick so we can
stop doing that.


cawallin:
looks good!

PR #160, opened 2016-03-18
Build the Docker images in the Makefile

ebd2:
We had an infrastructure failure where the yum repo wasn't available. This
meant that the docker build process for the base image for every test failed.
Unfortunately, we check for the base image and attempt to build it once per
testcase rather than once per test run, so our tests took forever to run and
had an error for every test. This commit moves building the base image to the
Makefile so that if we hit this issue in the future, we'll fail fast.


cawallin:
looks good

ebd2:
Thanks, I'll push after I make sure everything still works.

ebd2:
Thanks, I'll push after I make sure everything still works.

PR #161, opened 2016-03-18
Fix bug with collecting multiple server logs

rschlussel:
Testing: test_collect

@amrutagokhale @ebd2

### DIFF
--- a/prestoadmin/collect.py
+++ b/prestoadmin/collect.py
@@ -91,15 +91,15 @@ def get_remote_log_files(dest_path):
     remote_server_log = lookup_server_log_file(env.host)
     _LOGGER.debug('Logs to be archived on host ' + env.host + ': ' +
                   remote_server_log)
-    file_get(remote_server_log + '*', dest_path)
+    get_files(remote_server_log + '*', dest_path)
 
     remote_launcher_log = lookup_launcher_log_file(env.host)
     _LOGGER.debug('LOG directory to be archived on host ' + env.host + ': ' +
                   remote_launcher_log)
-    file_get(remote_launcher_log + '*', dest_path)
+    get_files(remote_launcher_log + '*', dest_path)
 
 
-def file_get(remote_path, local_path):
+def get_files(remote_path, local_path):
     path_with_host_name = os.path.join(local_path, env.host)
 
     if not os.path.exists(path_with_host_name):
@@ -107,9 +107,9 @@ def file_get(remote_path, local_path):
 
     _LOGGER.debug('local path used ' + path_with_host_name)
 
-    if exists(remote_path, True):
+    try:
         get(remote_path, path_with_host_name, True)
-    else:
+    except SystemExit:
         warn('remote path ' + remote_path + ' not found on ' + env.host)
 
 
@@ -227,7 +227,7 @@ def get_system_info(download_location):
 
     _LOGGER.debug('Gathered version information in file: ' + version_file_name)
 
-    file_get(version_file_name, download_location)
+    get_files(version_file_name, download_location)
 
 
 def get_platform_information():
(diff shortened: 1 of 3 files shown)


tests/product/test_collect.py
```diff
@@ -294,3 +274,33 @@ def test_collect_logs_server_not_installed(self):
                            self.PA_ONLY_CLUSTER)
         self.upload_topology()
         self._assert_no_logs_downloaded()
+
+    def test_collect_logs_multiple_server_logs(self):
+        self.setup_cluster(NoHadoopBareImageProvider(),
+                           self.STANDALONE_PRESTO_CLUSTER)
+        self.run_prestoadmin('server start')
+        self.cluster.write_content_to_host('#!/bin/bash\n touch '
+                                           '/var/log/presto/server.log-2',
```
ebd2: Please put touch on the next line so it's with its argument
Also, do we need to touch another log file to ensure that we fetch multiple logs if server.log doesn't exist? If not, please comment explaining why.

tests/product/test_collect.py
```diff
@@ -294,3 +274,33 @@ def test_collect_logs_server_not_installed(self):
                            self.PA_ONLY_CLUSTER)
         self.upload_topology()
         self._assert_no_logs_downloaded()
+
+    def test_collect_logs_multiple_server_logs(self):
+        self.setup_cluster(NoHadoopBareImageProvider(),
+                           self.STANDALONE_PRESTO_CLUSTER)
+        self.run_prestoadmin('server start')
+        self.cluster.write_content_to_host('#!/bin/bash\n touch '
```
ebd2: How about just write_content_to_host('fake log content', '/var/log/presto/server.log-2', self.cluster.master)

rschlussel: That works.  I was trying to write to all the hosts, but it actually only matters than one host has multiple server logs.

tests/product/test_collect.py
```diff
@@ -294,3 +274,33 @@ def test_collect_logs_server_not_installed(self):
...
+                               self.assertEqual, actual, expected)
+
+        downloaded_logs_location = path.join(TMP_PRESTO_DEBUG, 'logs')
+        self.assert_path_exists(self.cluster.master,
+                                downloaded_logs_location)
+
+        for host in self.cluster.all_internal_hosts():
+            host_log_location = path.join(downloaded_logs_location,
+                                          host)
+            self.assert_path_exists(self.cluster.master,
```
ebd2: I guess we're asserting that we get server.log. Maybe we should just touch it as well in case the code that gets us to this point changes in a way that we don't get a real server.log file?

ebd2:
LGTM % comments.

tests/product/test_collect.py
rschlussel: I think we'd want the tests to fail in that case because we probably need to change something with the collect command.

rschlussel:
updated

prestoadmin/collect.py
```diff
@@ -107,9 +107,9 @@ def file_get(remote_path, local_path):
 
     _LOGGER.debug('local path used ' + path_with_host_name)
 
-    if exists(remote_path, True):
+    try:
         get(remote_path, path_with_host_name, True)
-    else:
+    except SystemExit:
         warn('remote path ' + remote_path + ' not found on ' + env.host)
```
amrutagokhale: Can we change the name of this method to get_files to signify that there could be one or more files being fetched?

rschlussel: Done

amrutagokhale:
LGTM

tests/product/test_collect.py
ebd2: I agree with that line of thought. Thanks!

ebd2:
LGTM too!

rschlussel:
merged

PR #162, opened 2016-03-21
fixup! Add check to presto-admin that id_rsa is present

cawallin:
Because of error:
OSError: [Errno 2] ls: cannot access /home/app-admin/.ssh: No such file or directory


rschlussel:
lgtm

PR #163, opened 2016-03-21
Improve error message for assertRegexpMatches()

rschlussel:
Include the regex and text you are trying to match in the error message

Inspired by the automation failure that had no helpful output.  
@cawallin

### DIFF
--- a/tests/base_test_case.py
+++ b/tests/base_test_case.py
@@ -135,6 +135,7 @@ class BaseTestCase(unittest.TestCase):
     # equivalent to python 2.7's unittest.assertRegexpMatches()
     def assertRegexpMatches(
             self, text, expected_regexp, msg="Regexp didn't match"):
+        msg = '%s: %r not found in %r' % (msg, expected_regexp, text)
         self.assertTrue(re.search(expected_regexp, text), msg)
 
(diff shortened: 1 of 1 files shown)


cawallin:
looks good

tests/base_test_case.py
```diff
@@ -135,6 +135,7 @@ def assertLazyMessage(self, msg_func, assert_function, *args, **kwargs):
     # equivalent to python 2.7's unittest.assertRegexpMatches()
     def assertRegexpMatches(
             self, text, expected_regexp, msg="Regexp didn't match"):
+        msg = '%s: %r not found in %r' % (msg, expected_regexp, text)
```
ebd2: This is problematic for multi-line stuff. BaseMainCase._format_expected_actual might be borrow/adapt-able to this.

rschlussel: Pushed already. But i'll fix it.

ebd2: Thanks. Could you please fixe assertRaisesRegexp while you're in here? I think we can just replace the assertTrue(re.search(...)) with a call to assertRegexMaches. You can close SWARM-2005 too, by the way :-)

PR #164, opened 2016-03-21
Make regex not found error readable for long input

rschlussel:
Make the error message for assertRegexpMatches and assertRaisesRegex
more readable for input with newlines.

@ebd2


ebd2:
LGTM! Thanks!

PR #168, opened 2016-03-23
Add explicit argparse dependency.

ebd2:
Build a working offline dist if the system python includes argparse.


cawallin:
looks good

PR #171, opened 2016-03-29
Securely deploy config and connector files.

ebd2:
The configuration files and connector files we push to the presto coordinator and worker nodes can contain sensitive information. In particular, connector files can contain passwords. The ownership and permissions should be set such that only the presto user can access the files.

### DIFF
--- a/prestoadmin/connector.py
+++ b/prestoadmin/connector.py
@@ -19,12 +19,13 @@ import logging
 import errno
 
 from fabric.api import task, env
-from fabric.context_managers import hide
+from fabric.context_managers import hide, settings
 from fabric.contrib import files
-from fabric.operations import sudo, os, put, get
+from fabric.operations import sudo, os, put, get, abort
 import fabric.utils
 
-from prestoadmin.standalone.config import StandaloneConfig
+from prestoadmin.standalone.config import StandaloneConfig, \
+    PRESTO_STANDALONE_USER_GROUP
 from prestoadmin.util import constants
 from prestoadmin.util.base_config import requires_config
 from prestoadmin.util.exception import ConfigFileNotFoundError, \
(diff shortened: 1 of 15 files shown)


prestoadmin/connector.py
```diff
@@ -37,11 +38,30 @@
...
     _LOGGER.info('Deploying configurations for ' + str(filenames))
     sudo('mkdir -p ' + remote_dir)
     for name in filenames:
-        put(os.path.join(local_dir, name), remote_dir, use_sudo=True)
+        files = put(os.path.join(local_dir, name), remote_dir, use_sudo=True,
+                    mode=mode)
+        for file in files:
+            with settings(warn_only=True):
+                command = \
+                    "getent passwd {user} >/dev/null || ( rm -f {file} ; " \
```
cawallin: can you add commas around the || for clarity? I'm assuming that this check means that if the user doesn't exist, we fail to move the connector there, but it took a little while to parse.

prestoadmin/connector.py
```diff
@@ -37,11 +38,30 @@
...
+                    "exit {missing_owner_code} ) && " \
+                    "chown {user_group} {file}".format(
+                        user=user, file=file, user_group=user_group,
+                        missing_owner_code=missing_owner_code)
+
+                result = sudo(command)
+
+                if result.return_code == missing_owner_code:
+                    abort("User %s does not exist. Make sure the presto "
```
cawallin: s/presto/Presto

cawallin:
looks good % a few comments and upcoming tests

prestoadmin/connector.py
ebd2: Fixed here and elsewhere.

prestoadmin/connector.py
ebd2: ditto.

cawallin:
Could you add a test for if the presto user isn't there?
You should also add release notes about this.
Otherwise looks good.

ebd2:
@cawallin: Added the requested tests, fixed other busted product tests. Review notes forthcoming.

cawallin:
looks good, though you should hold off on pushing it to the release branch

PR #174, opened 2016-04-05
Use sudo when reading config.properties during port lookup

anusudarsan:
Presto configuration files are  only readable by "presto" user.
So when presto-admin user is not "presto", config.properties file read
(for port lookup) during "server status" fails. This patch
fixes the issue by using sudo for config file read.

Testing: Tested in EMR where presto-admin user is hadoop, make clean
lint test-all

### DIFF
--- a/prestoadmin/util/remote_config_util.py
+++ b/prestoadmin/util/remote_config_util.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 import logging
 from fabric.context_managers import settings, hide
-from fabric.operations import run
+from fabric.operations import sudo
 from fabric.tasks import execute
 from prestoadmin.util.exception import ConfigurationError
 from prestoadmin.util.constants import DEFAULT_PRESTO_LAUNCHER_LOG_FILE,\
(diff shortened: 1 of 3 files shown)


anusudarsan:
@cawallin @ebd2 Can you review please ?

prestoadmin/util/remote_config_util.py
```diff
@@ -81,7 +81,7 @@ def lookup_string_config(config_value, config_file, host, default=''):
 
 def lookup_in_config(config_key, config_file, host):
     with settings(hide('stdout', 'warnings', 'aborts')):
-        config_value = execute(run, 'grep %s= %s' % (config_key, config_file),
+        config_value = execute(sudo, 'grep %s= %s' % (config_key, config_file),
```
ebd2: Should we do sudo -u presto instead? I don't feel strongly though.

cawallin: eh, this is using fabric's sudo method and I'm not sure whether you can change the user?

cawallin:
looks good

prestoadmin/util/remote_config_util.py
anusudarsan: there are "user" and "group" arguments we can pass to this sudo method. Do we want to use that?

cawallin: As long as this works, I'm not sure it matters

anusudarsan:
updated and tested it works in EMR

prestoadmin/util/remote_config_util.py
cawallin: Ah, I see you changed it, that's fine too

ebd2: I'm not sure it matters in practice either, but it does limit the permissions you need to give somebody.

ebd2:
Forgot to mention earlier: LGTM either way :-)

anusudarsan:
Ya, I updated with user and group args to the sudo method. verified it works in EMR

PR #175, opened 2016-04-11
Make file_content_message output more clear

ebd2:
Include actual, expected, and presto-admin file content between explicit
begin and end markers.

### DIFF
--- a/tests/product/base_product_case.py
+++ b/tests/product/base_product_case.py
@@ -332,6 +332,11 @@ query.max-memory=50GB\n"""
               '' % (actual, expected)
         if pa_file:
             try:
+                # If the actual file content should have come from a file that
+                # lives on the presto-admin host that we shove over to some
+                # other host, display the content of the file as it is on the
+                # presto-admin host. Presumably this will match the actual
+                # file content that we display above.
                 msg += '\t==== Content for presto-admin file %s ====\n' % \
                        (pa_file,)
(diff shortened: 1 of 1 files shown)


tests/product/base_product_case.py
```diff
@@ -324,12 +324,19 @@ def assert_file_content(self, host, filepath, expected):
...
+              '\t=========== DID NOT EQUAL ===========\n' \
+              '%s\n' \
+              '\t==== ^^ EXPECTED FILE CONTENT ^^ ====\n' \
+              '' % (actual, expected)
         if pa_file:
             try:
-                msg += '\n Content for presto-admin file %s \n' % pa_file
+                msg += '\t==== Content for presto-admin file %s ====\n' % \
+                       (pa_file,)
                 msg += self.get_file_content(self.cluster.get_master(),
```
cawallin: this should also be against the correct host

ebd2: Sorry, this is confusing and I should have added a comment earlier. There's one now. TL;DR: the master is the correct host, we're getting the file as it exists on the presto-admin host as a sanity check of the actual file content.

cawallin: I was confused by this because I thought it was saying that the contents of the file were changing when checking it a second time, which would suggest an intermittent issue.

cawallin:
looks good

ebd2:
Squashed commits and pushed.

PR #177, opened 2016-05-12
Unify cluster bringup from bare images.

ebd2:
test_installer and prestoadmin_installer both did a bunch of low-level
interaction with docker_cluster to bring up a bare cluster for their various
purposes. We already have code that handles bringing up a bare cluster that
base_product_test uses.

This reworks the bare cluster bringup code so that it can be used by
test_installer and prestoadmin_installer and gets them out of the business of
creating their own bare images.


ebd2:
@cawallin, @petroav, @rschlussel

tests/product/constants.py
```diff
@@ -27,8 +27,4 @@
 DEFAULT_DOCKER_MOUNT_POINT = '/mnt/presto-admin'
 DEFAULT_LOCAL_MOUNT_POINT = os.path.join(main_dir, 'tmp/docker-pa/')
 
-BASE_IMAGE_NAME = 'jdeathe/centos-ssh'
-BASE_IMAGE_TAG = 'centos-6-1.2.0'
-
 BASE_TD_IMAGE_NAME = 'teradatalabs/centos6-ssh-test'
```
cawallin: is this used anywhere? this needs to change

ebd2: It needs to go away, but it's going to happen in the upcoming "make  _ensure_docker_containers_started opt-out rather than opt-in" PR. For the moment _ensure... still refers to it.

tests/product/test_installer.py
```diff
@@ -59,25 +57,26 @@ def test_offline_installer(self):
...
-            cap_add=['NET_ADMIN']
-        )
+        cluster_type = 'installer_tester'
+        bare_image_provider = NoHadoopBareImageProvider()
+        centos_container, bare_cluster = DockerCluster.start_cluster(
+            bare_image_provider, cluster_type, 'master', [],
+            cap_add=['NET_ADMIN'])
+
+        if bare_cluster:
+            centos_container.commit_images(
```
cawallin: I'm not convinced we have to commit this image -- do we clean it up anywhere?

tests/product/base_product_case.py
```diff
@@ -170,14 +170,12 @@ def setup_cluster(self, bare_image_provider, cluster_type):
                 StandalonePrestoInstaller.assert_installed)
         else:
             try:
-                self.cluster = DockerCluster.start_existing_images(
+                self.cluster, bare_cluster = DockerCluster.start_cluster(
                     bare_image_provider, cluster_type)
-                if self.cluster:
+                if not bare_cluster:
```
cawallin: why "if not bare_cluster" here? don't you always want to run the installers?

ebd2: Nope. It's a little weird. I'll add a comment.

cawallin:
lgtm

tests/product/test_installer.py
ebd2: We don't need to because we don't change anything on container before committing them, but it does keep everything consistent from a usage standpoint. Also, they don't take up any space because layers. The same is true for `installer_builder` in prestoadmin_installer.py

`make clean-test` removes all of the images under teradatalabs/pa_test, so they will get cleaned up.

cawallin: ok cool, feel free to keep this

ebd2:
Squashed and closed.

PR #182, opened 2016-06-09
Clarify package upgrade and error if bad input

cawallin:
* Update docs to make it clear that you need to specify the RPM,
not the directory
* Error out if you specify a directory in which an RPM is,
rather than the actual location of the RPM

### DIFF
--- a/docs/presto-admin-commands.rst
+++ b/docs/presto-admin-commands.rst
@@ -489,15 +489,16 @@ server upgrade
 **************
 ::
 
-    presto-admin server upgrade local_package_path [local_config_dir]
+    presto-admin server upgrade path/to/new/package.rpm [local_config_dir]
 
-This command upgrades the Presto RPM on all of the nodes in the cluster to the RPM specified
-by ``local_package_path``, preserving the existing configuration on the cluster. The existing
+This command upgrades the Presto RPM on all of the nodes in the cluster to the RPM at
+``path/to/new/package.rpm``, preserving the existing configuration on the cluster. The existing
 cluster configuration is saved locally to local_config_dir (which defaults to a temporary
-folder if not specified).
+folder if not specified). The path can either be absolute or relative to the current
+directory.
 
-This command can also be used to downgrade the Presto installation, if the RPM at ``local_package_path``
-is an earlier version than the Presto installed on the cluster.
+This command can also be used to downgrade the Presto installation, if the RPM at
+``path/to/new/package.rpm`` is an earlier version than the Presto installed on the cluster.
 
 Note that if the configuration files on the cluster differ from the presto-admin configuration
 files found in ``/etc/opt/prestoadmin``, the presto-admin configuration files are not updated.
@@ -506,7 +507,8 @@ Example
 -------
 ::
 
-    sudo ./presto-admin server upgrade new-rpm.rpm /tmp/cluster-configuration
+    sudo ./presto-admin server upgrade path/to/new/package.rpm /tmp/cluster-configuration
+    sudo ./presto-admin server upgrade /path/to/new/package.rpm /tmp/cluster-configuration
 
 
 *************
--- a/prestoadmin/package.py
+++ b/prestoadmin/package.py
@@ -74,6 +74,9 @@ def deploy_action(local_path, rpm_action):
 
 
 def deploy(local_path=None):
+    if not os.path.isfile(local_path):
+        abort('RPM file not found at %s.' % local_path)
+
     _LOGGER.info("Deploying rpm on %s..." % env.host)
     print("Deploying rpm on %s..." % env.host)
     sudo('mkdir -p ' + constants.REMOTE_PACKAGES_PATH)
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -166,7 +166,7 @@ def uninstall():
 
 @task
 @requires_config(StandaloneConfig)
-def upgrade(local_package_path, local_config_dir=None, overwrite=False):
+def upgrade(new_rpm_path, local_config_dir=None, overwrite=False):
     """
     Copy and upgrade a new presto-server rpm to all of the nodes in the
     cluster. Retains existing node configuration.
@@ -183,8 +183,7 @@ def upgrade(local_package_path, local_config_dir=None, overwrite=False):
     Note that the configuration files in /etc/opt/prestoadmin are not updated
     during upgrade.
 
-    :param local_package_path - Absolute path to the presto rpm to be
-                                installed
+    :param new_rpm_path -       The path to the new Presto RPM to install
     :param local_config_dir -   (optional) Directory to store the cluster
                                 configuration in. If not specified, a temp
                                 directory is used.
@@ -200,7 +199,7 @@ def upgrade(local_package_path, local_config_dir=None, overwrite=False):
     configure_cmds.gather_directory(local_config_dir, overwrite)
     filenames = connector.gather_connectors(local_config_dir, overwrite)
 
-    package.deploy_upgrade(local_package_path)
+    package.deploy_upgrade(new_rpm_path)
 
     configure_cmds.deploy_all(local_config_dir)
     connector.deploy_files(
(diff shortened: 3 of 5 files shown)


docs/presto-admin-commands.rst
```diff
@@ -489,14 +489,14 @@ server upgrade
 **************
 ::
 
-    presto-admin server upgrade local_package_path [local_config_dir]
+    presto-admin server upgrade new-rpm [local_config_dir]
```
maciejgrzybek: `path_to_new_rpm` maybe?

docs/presto-admin-commands.rst
```diff
@@ -489,14 +489,14 @@ server upgrade
...
 
-This command upgrades the Presto RPM on all of the nodes in the cluster to the RPM specified
-by ``local_package_path``, preserving the existing configuration on the cluster. The existing
+This command upgrades the Presto RPM on all of the nodes in the cluster to the RPM
+``new-rpm``, preserving the existing configuration on the cluster. The existing
```
maciejgrzybek: Maybe clarify if given path is relative to `presto-admin` binaries or it has to be absolute?

docs/presto-admin-commands.rst
```diff
@@ -506,7 +506,7 @@ Example
 -------
 ::
 
-    sudo ./presto-admin server upgrade new-rpm.rpm /tmp/cluster-configuration
+    sudo ./presto-admin server upgrade /path/to/new-rpm.rpm /tmp/cluster-configuration
```
maciejgrzybek: This should be corresponding to however we describe it here: https://github.com/prestodb/presto-admin/pull/182/files#r66441884.
If we describe it as absolute path, it's fine how's it now. If we are going to refer to it as a relative path, it should be relative in the example too.

docs/presto-admin-commands.rst
cawallin: I think the whole issue that John had was with the word path -- it made him think that he should specify the directory

tests/product/test_server_upgrade.py
```diff
@@ -107,3 +107,28 @@ def test_rolling_upgrade(self):
...
+            'server upgrade ' + dir_on_cluster
+        )
+
+    def test_upgrade_works_with_symlink(self):
+        self.start_and_assert_started()
+
+        self.run_prestoadmin('configuration deploy')
+        for container in self.cluster.all_hosts():
+            self.real_installer.assert_installed(self, container)
+            self.assert_has_default_config(container)
```
maciejgrzybek: Shouldn't we assert some different configuration?
Even if upgrade overwrites configuration, it would overwrite it to default, so this test does not check if we're preserving configuration on upgrade.

maciejgrzybek: Maybe it should be a separate test though. E.g. `test_upgrade_preserves_configuration`.

docs/presto-admin-commands.rst
maciejgrzybek: What if you put `path/to/new/package.rpm`. Isn't it clear that you have to provide path with filename?

ebd2: I second maciej's suggestion of path/to/new/package.rpm. It makes it clear that it it's a path to an rpm file.

docs/presto-admin-commands.rst
ebd2: Can't it be either?

prestoadmin/server.py
```diff
@@ -183,8 +183,7 @@ def upgrade(local_package_path, local_config_dir=None, overwrite=False):
     Note that the configuration files in /etc/opt/prestoadmin are not updated
     during upgrade.
 
-    :param local_package_path - Absolute path to the presto rpm to be
-                                installed
+    :param new_rpm -            The new Presto RPM to be installed
```
ebd2: How about "The path to the new Presto RPM to install"?

prestoadmin/package.py
```diff
@@ -74,6 +74,10 @@ def deploy_action(local_path, rpm_action):
 
 
 def deploy(local_path=None):
+    if not os.path.isfile(local_path):
+        abort('RPM file not found at %s. Make sure to specify the filename'
+              'and not the directory name.' % local_path)
```
ebd2: This seem like an awfully specific error message.

prestoadmin/package.py
```diff
@@ -74,6 +74,10 @@ def deploy_action(local_path, rpm_action):
 
 
 def deploy(local_path=None):
+    if not os.path.isfile(local_path):
```
ebd2: Should we verify that it's readable too? That seems like a common enough case that it shouldn't result in a stack trace for the user.

prestoadmin/server.py
```diff
@@ -166,7 +166,7 @@ def uninstall():
 
 @task
 @requires_config(StandaloneConfig)
-def upgrade(local_package_path, local_config_dir=None, overwrite=False):
+def upgrade(new_rpm, local_config_dir=None, overwrite=False):
```
ebd2: How about new_rpm_path?

docs/presto-admin-commands.rst
maciejgrzybek: Okay, let me rephrase it:
if we expect absolute path (e.g. because it's got to be available on all nodes in the same location or for whatever else reason), let's make it clear here.

tests/product/test_server_upgrade.py
cawallin: good point. a little out of the scope of this PR, but I'll do it :)

prestoadmin/package.py
cawallin: You get a pretty helpful error message in put about this:
Fatal error: [hdp-kerb3-master] error: jre-8u77-linux-x64.rpm: open failed: Permission denied

That's probably more specific than what I would give

cawallin:
Updated

prestoadmin/package.py
ebd2: Sounds reasonable; I'm OK leaving it as-is

prestoadmin/server.py
```diff
@@ -183,8 +183,8 @@ def upgrade(local_package_path, local_config_dir=None, overwrite=False):
     Note that the configuration files in /etc/opt/prestoadmin are not updated
     during upgrade.
 
-    :param local_package_path - Absolute path to the presto rpm to be
-                                installed
+    :param new_rpm_path -       The path to the new Presto Presto RPM to
```
ebd2: "Presto Presto"

tests/product/test_server_upgrade.py
```diff
@@ -107,3 +107,68 @@ def test_rolling_upgrade(self):
...
+            self.assert_has_default_connector(container)
+
+        path_on_cluster = self.copy_upgrade_rpm_to_cluster()
+        symlink = '/opt/prestoadmin/link.rpm'
+        self.cluster.exec_cmd_on_host(self.cluster.master, 'ln -s %s %s'
+                                      % (path_on_cluster, symlink))
+        self.upgrade_and_assert_success(symlink)
+
+    def test_configuration_preserved_on_upgrade(self):
```
ebd2: Nice. Thanks!

tests/product/test_server_upgrade.py
```diff
@@ -107,3 +107,68 @@ def test_rolling_upgrade(self):
...
+        symlink = '/opt/prestoadmin/link.rpm'
+        self.cluster.exec_cmd_on_host(self.cluster.master, 'ln -s %s %s'
+                                      % (path_on_cluster, symlink))
+
+        self.run_prestoadmin('server upgrade ' + path_on_cluster)
+        self.assert_dummy_properties(self.cluster.slaves[1])
+
+    def add_dummy_properties_to_host(self, host):
+        self.cluster.write_content_to_host(
+            'com.facebook.presto=INFO',
```
ebd2: Extract constants here and below?

ebd2:
LGTM mod the typo and pulling out constants if you're inclined to do so.

maciejgrzybek:
me to goods look

PR #187, opened 2016-06-21
Correct port-in-use test

ebd2:
Don't match port if it isn't a whole word. I.e. don't match on 18080.
Testing:
```
# nc -l 18080 
# presto-admin server start
```
Make sure presto starts cleanly.

### DIFF
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -242,7 +242,7 @@ def is_port_in_use(host):
                      "Skipping check for port already being used")
         return 0
     with settings(hide('warnings', 'stdout'), warn_only=True):
-        output = run('netstat -an |grep %s |grep LISTEN' % str(portnum))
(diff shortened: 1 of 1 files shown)


rschlussel:
lgtm

cawallin:
can you add a test that would fail if the port check is wrong?

ebd2:
I thought about how I would do this. Here's what I was thinking:

1. Install netcat on the docker cluster
2. Run netcat -l 18080.
3. netcat doesn't exit, so either do it in another thread or figure out if netcat has a daemon mode?
4. Server start
5. Scrape output/etc
6. Kill off another thread if that's what we ended up doing?

Since there's a low risk of anybody ever wanting to change this, I'm not overly inclined to write a product test for it. The code has been unchanged since it was originally written a year ago.

cawallin:
Fair enough, that's ok.

PR #188, opened 2016-06-24
Use JAVA8_HOME for server upgrade

ebd2:
Tested manually. Adding a product test would require depending on the dummy upgrade rpm to have an up to date version of the java detection logic, which it presently does not.

If we want to add a product test for this, we need to come up with a way to ensure the dummy rpm reflects the current state of presto.


cawallin:
looks good

ebd2:
Rebased and pushed

PR #193, opened 2016-06-29
Add --nodeps option to server uninstall

anusudarsan:
This commit adds a --nodeps optional flag to server uninstall. The
functionality is similar to --nodeps flag to rpm -e.

Ideally, we want the --nodeps optional flag to be present for all
rpm operations, like server install and server upgrade. But that will
be supported later.

Testing: manual testing, make clean lint test

### DIFF
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -107,7 +107,12 @@ def install(local_path):
     jvm.config
 
     Parameters:
-        local_path - Absolute path to the presto rpm to be installed
+        local_path -            Absolute path to the presto rpm to be installed
+
+        --nodeps -              (optional) Flag to indicate if server install
+                                should ignore checking Presto rpm package
+                                dependencies. Equivalent to adding --nodeps
+                                flag to rpm -i.
     """
     package.check_if_valid_rpm(local_path)
     return execute(deploy_install_configure, local_path, hosts=get_host_list())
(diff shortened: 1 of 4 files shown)


anusudarsan:
@petroav Can you review please?

prestoadmin/server.py
```diff
@@ -149,17 +149,26 @@ def wait_for_presto_user():
 def uninstall():
     """
     Uninstall Presto after stopping the services on all nodes
+
+    Args:
+        --nodeps (optional): Flag to indicate if server uninstall
+        should ignore checking Presto rpm package dependencies.
+        Equivalent to adding --nodeps flag to rpm -e
```
petroav: Full stop at the end.

prestoadmin/server.py
```diff
@@ -149,17 +149,26 @@ def wait_for_presto_user():
...
     stop()
 
+    nodeps = ''
+    if env.nodeps:
+        nodeps = ' --nodeps'
+
     # currently we have two rpm names out so we need this retry
     with quiet():
-        ret = sudo('rpm -e presto')
+        ret = sudo('rpm -e%s presto' % (nodeps))
```
petroav: I would rather include the the space as part of the formatting string instead of as part of variable that gets substituted in.

rschlussel:
I think the --nodeps flag should be added for server install and server upgrade as part of this.  upgrade just does uninstall and then install, so you want that to be able to have --nodeps too.  And even though it wouldn't currently make any practical difference with install, there's no reason to introduce an asymmetry.

rschlussel:
I would say even that it's essential for upgrade, because otherwise the upgrade can fail even though you're about to replace the package.

anusudarsan:
To add it to uninstall came up as a requirement specifically. See https://jira.td.teradata.com/jira/browse/SWARM-2593. Looks like this was needed for Q2 (tomorrow?). So I thought of adding to uninstall for now and do the server install and upgrade as a follow-on patch. But I agree it is better to add it everywhere. @petroav Is it ok if we merge this on Friday?

petroav:
I didn't realize that upgrade was remove followed by install. This means that if you wanted to upgrade just Presto without upgrading QG, you wouldn't be able to because on remove the operation would fail. This is a good argument I didn't think of, so let's add --nodeps to install and upgrade.

We haven't created a tag or release branch for presto-admin so this can definitely go in on Friday. Even if we create a release branch/tag tomorrow we can always cherry pick the commit over.

anusudarsan:
Cool. I will apply the same to install and upgrade.

anusudarsan:
Updated. The --nodeps functionality for upgrade and install was already present. I just added the help text and updated the documentation.

rschlussel:
lgtm % Anton's comment about the formatting string

anusudarsan:
re: formatting string I was getting weird unit test failures when using the format suggested. The unit test for uninstall without --nodeps was failing with an error that `sudo(rpm -e presto) is not called`. When I debug the string looked fine (no extra whitespaces). I looked at package.install --nodeps and it has the same formatting as I have. So keeping it as-is.

rschlussel:
lgtm

PR #195, opened 2016-06-30
Extract building dist package out of prestoadmin_installer

### LINKED ISSUE #189: Extract package build from product tests
Code under https://github.com/prestodb/presto-admin/blob/master/tests/product/prestoadmin_installer.py#L80 should be extracted as some `make build` command. Thanks to that user (developer) has a control when a package is rebuilt. 

Currently it rebuilds automatically, which is not obvious. I didn't know about that and I took me a some time to investigate how presto admin package is installed on docker.

kokosing:
Extract building dist package out of prestoadmin_installer

Build binary files seems to be a better task for the build tool (like make) than
for the product tests framework. Moreover product tests framework was
caching the result in the `tmp/installer` directory so when the changes
were introduced to presto-admin, package was not rebuild. So it was easy
to run product tests with the old version of product-tests.

Instead of that building distribution package was moved to Makefile.
Moreover, disregarding how this file is created it will be stored in
`dist` directory. No more need for having `tmp/installer`.

Fixes #189

### DIFF
--- a/.travis.yml.disabled
+++ b/.travis.yml.disabled
@@ -18,7 +18,7 @@ env:
     - PRODUCT_TEST_GROUP=6
     - PRODUCT_TEST_GROUP=7
 before_script:
-  - pip install --upgrade pip==6.1.1
+  - pip install --upgrade pip==7.1.2
   - pip install -r requirements.txt
   - make docker-images
 install: 
--- a/.travis.yml.disabled
+++ b/.travis.yml.disabled
@@ -17,11 +17,12 @@ env:
     - PRODUCT_TEST_GROUP=5
     - PRODUCT_TEST_GROUP=6
     - PRODUCT_TEST_GROUP=7
-install: 
+before_script:
   - pip install --upgrade pip==6.1.1
   - pip install -r requirements.txt
-before_script:
   - make docker-images
+install: 
+  - make docker-dist-online
 script: 
   - |
     if [ -v PRODUCT_TEST_GROUP ]; then
--- a/Makefile
+++ b/Makefile
@@ -83,7 +83,7 @@ test: clean-test
 
 TEST_SUITE?=tests.product
 
-test-all: clean-test docker-images
+test-all: clean-test docker-images docker-dist
 	tox -- -s tests.unit
 	tox -- -s tests.integration
 	tox -e py26 -- -s ${TEST_SUITE} -a '!quarantine'
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 .PHONY: clean-all clean clean-eggs clean-build clean-pyc clean-test-containers clean-test \
 	clean-docs lint smoke test test-all test-rpm coverage docs open-docs release release-builds \
-	dist dist-online dist-offline wheel install precommit
+	dist dist-online dist-offline wheel install precommit \
+	docker-dist docker-dist-online docker-dist-offline
 
 help:
 	@echo "precommit - run \`quick' tests and tasks that should pass or succeed prior to pushing"
@@ -25,6 +26,9 @@ help:
 	@echo "dist - package and build installer that requires an Internet connection"
 	@echo "dist-online - package and build installer that requires an Internet connection"
 	@echo "dist-offline - package and build installer that does not require an Internet connection"
+	@echo "docker-dist - package and build installer (within a docker container) that requires an Internet connection"
+	@echo "docker-dist-online - package and build installer (within a docker container) that requires an Internet connection"
+	@echo "docker-dist-offline - package and build installer (within a docker container) that does not require an Internet connection"
 	@echo "wheel - build wheel only"
 	@echo "install - install the package to the active Python's site-packages"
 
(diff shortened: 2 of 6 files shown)


.travis.yml.disabled
```diff
@@ -17,11 +17,12 @@ env:
     - PRODUCT_TEST_GROUP=5
     - PRODUCT_TEST_GROUP=6
     - PRODUCT_TEST_GROUP=7
-install: 
+before_script:
   - pip install --upgrade pip==6.1.1
```
rschlussel: not related to this change, but the pip version in requirements.txt is 7.1.2  (and maybe should be bumped up to the latest version of pip)

Makefile
```diff
@@ -121,6 +125,14 @@ dist-offline: clean-build clean-pyc
 	python setup.py bdist_prestoadmin
 	ls -l dist
 
+docker-dist: docker-dist-online
+
+docker-dist-online: clean-build clean-pyc
```
rschlussel: these should get called at the beginning of make test-all

tests/product/prestoadmin_installer.py
```diff
@@ -69,84 +57,16 @@ def assert_installed(testcase, msg=None):
...
-    def _copy_dist_to_host(cluster, local_dist_dir, dest_host):
-        for dist_file in os.listdir(local_dist_dir):
+    def copy_dist_to_master(self, cluster, dist_dir):
+        if (not os.path.isdir(dist_dir) or
+                len(fnmatch.filter(
+                    os.listdir(dist_dir),
+                    'prestoadmin-*.tar.bz2')) == 0):
+            self.testcase.fail(
+                'Unable to find presto-admin package. Have you run one of '
+                '`make dist*` or `make docker-dist*` command?')
```
rschlussel: you can get rid of the word 'command' at the end

tests/product/prestoadmin_installer.py
```diff
@@ -69,84 +57,16 @@ def assert_installed(testcase, msg=None):
...
+        for dist_file in os.listdir(dist_dir):
             if fnmatch.fnmatch(dist_file, "prestoadmin-*.tar.bz2"):
                 cluster.copy_to_host(
-                    os.path.join(local_dist_dir, dist_file),
-                    dest_host)
+                    os.path.join(dist_dir, dist_file),
```
rschlussel: what happens if there is no match?

tests/product/test_installer.py
```diff
@@ -42,30 +42,12 @@ def tearDown(self):
         self.centos_container.tear_down()
 
     @attr('smoketest')
-    def test_online_installer(self):
-        self.pa_installer._build_installer_in_docker(self.centos_container,
-                                                     online_installer=True,
-                                                     unique=True)
+    def test_installer(self):
```
rschlussel: I'm confused about how this test works now.  How does the installer get into the docker container?

tests/product/test_installer.py
```diff
@@ -42,30 +42,12 @@ def tearDown(self):
...
         self.__verify_third_party_dir(False)
         self.pa_installer.install(
             dist_dir=self.centos_container.get_dist_dir(unique=True))
         self.run_prestoadmin('--help', raise_error=True)
 
-    @attr('smoketest', 'offline_installer')
-    def test_offline_installer(self):
```
rschlussel: why did you remove this?

tests/product/prestoadmin_installer.py
kokosing: above this `if` block there is a check if there is a file in this directory and fails if there isn't

tests/product/test_installer.py
kokosing: `prestoadmin_installer` takes it from `dist` directory. `dist` directory is a place where all `make dist dist-online dist-offline docker-dist docker-dist-online docker-distoffline` put result artifact. Previously `prestoadmin_installer` took that from `tmp/installer` which was akward to me.

tests/product/test_installer.py
kokosing: I wanted to make tests agnostic about what version (offline vs online) they tests.
My idea of having tests for online and offline was to run something like:

```
make dist-online
make test-all
```
and 
```
make dist-offline
make test-all
```

Now I noticed that `test_offline_installer` disables network in the container. I will revert it. But I remove it from the group `smoketests`, so that test should be run only with offline installer.

kokosing:
I applied the comments, but I am not going to push it until we have travis working so I am sure I didn't break anything.

tests/product/test_installer.py
rschlussel: I think it should still be run in automation, but it's fine if it's not in smoketest. 

Also, which gets run if you have both the online and offline installer in the dist directory? That's something we need to be careful of when running the tests.

petroav:
@kokosing just an FYI that for any changes to Travis presto-admin configuration make sure to talk to @zachyee who has been working on separating out tests into groups and is investigating how to fix the '10 minute no output' problem.

Makefile
```diff
@@ -121,6 +125,14 @@ dist-offline: clean-build clean-pyc
...
+	docker run --rm -v `pwd`:/workdir -w /workdir -u `id -u`:`id -g` teradatalabs/presto-admin-devenv:1 make dist-online
```
ebd2: Can you please update the readme to cover the following:

1. When you'd want to run dist-<online|offline> vs docker-dist<online|offline>
2. What python gets used to build the installers when you make any of the above targets
3. Where the installer ends up.

Thanks!

tests/product/test_installer.py
ebd2: How about adding targets
```
test-online: dist-online test-all
test-offline: dist-offline test-all
```
to cover those common use cases?

ebd2:
This is an awesome clean-up! Thanks so much for taking care of this.

I am delighted to see all the magic in prestoadmin_installer.py being removed.

ebd2:
Travis is working-ish; can you please rebase this after we release presto-admin so we can get this in?

kokosing:
I am going to rebase this pull request, so far I extracted this: https://github.com/prestodb/presto-admin/pull/320

PR #196, opened 2016-06-30
Add 'package uninstall' support

kokosing:
Add 'package uninstall' support

### DIFF
--- a/docs/presto-admin-commands.rst
+++ b/docs/presto-admin-commands.rst
@@ -319,6 +319,31 @@ Example
     sudo ./presto-admin package install /tmp/jdk-8u45-linux-x64.rpm
 
 
+*****************
+package uninstall
+*****************
+
+::
+
+    presto-admin package uninstall rpm_package_name [--nodeps]
+
+This command uninstalls an rpm package from all the nodes in the cluster. Similar to ``server uninstall`` the cluster
+topology is obtained from the file ``/etc/opt/prestoadmin/config.json``. If this file is missing, then the command
+prompts for user input to get the topology information.
+
+This command takes an optional ``--nodeps`` flag which indicates if the rpm installed should ignore checking any package
+dependencies.
+
+.. WARNING:: Using ``--nodeps`` can result in uninstalling the rpm even when dependant packages are installed. It may
+end up with a broken rpm installation.
+
+Example
+-------
+::
+
+    sudo ./presto-admin package uninstall jdk
+
+
 **************
 plugin add_jar
 **************
--- a/prestoadmin/package.py
+++ b/prestoadmin/package.py
@@ -30,7 +30,7 @@ from prestoadmin.util.base_config import requires_config
 from prestoadmin.util.fabricapi import get_host_list
 
 _LOGGER = logging.getLogger(__name__)
-__all__ = ['install']
+__all__ = ['install', 'uninstall']
 
 
 @task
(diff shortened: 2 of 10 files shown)


kokosing:
I have run limited amount of testing for that. Something like:
```
flake ...
make test
make smoke
```

Note that there is no automation enabled for pull requests, so there is a risk that this pull requests introduces some regression.

tests/unit/test_package.py
```diff
@@ -51,39 +51,22 @@ def test_rpm_install_nodeps(self, mock_sudo):
         mock_sudo.assert_called_with('rpm -i --nodeps '
                                      '/opt/prestoadmin/packages/test.rpm')
 
+    @patch('prestoadmin.package._rpm_uninstall')
+    @patch('prestoadmin.package._rpm_install')
     @patch('prestoadmin.package.sudo')
-    def test_rpm_upgrade(self, mock_sudo):
+    def test_rpm_upgrade(self, mock_sudo, mock_rpm_install, mock_rpm_uninstal):
```
rschlussel: missing an 'l' at the end of uninstall

prestoadmin/package.py
```diff
@@ -90,39 +90,72 @@ def deploy(local_path=None):
...
 
 
 def rpm_install(rpm_name):
     _LOGGER.info("Installing the rpm")
-    package_path = os.path.join(constants.REMOTE_PACKAGES_PATH, rpm_name)
-    ret = _rpm_install(package_path)
-    if ret.succeeded:
-        print("Package installed successfully on: " + env.host)
+    _rpm_install(_rpm_path(rpm_name))
+    print("Package installed successfully on: " + env.host)
```
rschlussel: This will always get printed because you removed the check that the installation was successful.  The return value from sudo tells you whether the command had a non-zero exit code.

docs/presto-admin-commands.rst
```diff
@@ -319,6 +319,27 @@ Example
...
 
+*****************
+package uninstall
+*****************
+
+::
+
+    presto-admin package uninstall rpm_package_name [--nodeps]
+
+This command uninstalls rpm package from all the nodes in the cluster. Similar to ``server uninstall`` the cluster topology is obtained from the file ``/etc/opt/prestoadmin/config.json``. If this file is missing, then the command prompts for user input to get the topology information.
```
rschlussel: s/uninstalls rpm package/uninstalls an rpm package

prestoadmin/package.py
```diff
@@ -90,39 +90,72 @@ def deploy(local_path=None):
...
+    _rpm_install(rpm_path)
+    print("Package upgraded successfully on: " + env.host)
+
+
+@task
+@runs_once
+@requires_config(StandaloneConfig)
+def uninstall(rpm_name):
+    """
+    Install the rpm package on the cluster
```
rschlussel: This should be:
Uninstall an rpm package from the cluster

prestoadmin/package.py
```diff
@@ -90,39 +90,72 @@ def deploy(local_path=None):
...
+            to adding --nodeps flag to rpm -e.
+    """
+    return execute(rpm_uninstall, rpm_name, hosts=get_host_list())
+
+
+def rpm_uninstall(package_name):
+    _LOGGER.info("Uninstalling the rpm")
+
+    _rpm_uninstall(package_name)
+    print("Package uninstalled successfully on: " + env.host)
```
rschlussel: you don't check the return value

prestoadmin/package.py
```diff
@@ -90,39 +90,72 @@ def deploy(local_path=None):
...
-    package_name = sudo('rpm -qp --queryformat \'%%{NAME}\' %s'
-                        % package_path, quiet=True)
 
-    ret_uninstall = sudo('rpm -e %s%s' % (nodeps, package_name))
-    ret_install = _rpm_install(package_path)
-    if ret_uninstall.succeeded and ret_install.succeeded:
-        print("Package upgraded successfully on: " + env.host)
+def _rpm_uninstall(package_name):
+    return sudo(
+        'if rpm -qi %(package)s > /dev/null; then'
```
rschlussel: This swallows the case where the package does not exist.  package install will give you an error if the package is already installed.  We should be consistent.  
Something that's a bit confusing about having it this way is that it prints out at the end that the package was successfully uninstalled, when nothing happened. if you have a typo in the rpm name, it will hide that from you

On the other hand there's an issue filed to allow it not to give an error, because when things are scripted it's annoying to have a non-zero exit code when nothing went wrong.

rschlussel:
you should write product tests.  and also run the product tests for package install and server install/uninstall

prestoadmin/package.py
```diff
@@ -90,39 +90,72 @@ def deploy(local_path=None):
...
+    Args:
+        rpm_name: Name of the rpm to be uninstalled
+        --nodeps (optional): Flag to indicate if rpm install (or uninstall)
+            should ignore checking package dependencies. Equivalent
+            to adding --nodeps flag to rpm -e.
+    """
+    return execute(rpm_uninstall, rpm_name, hosts=get_host_list())
+
+
+def rpm_uninstall(package_name):
```
rschlussel: you should change server uninstall to call this.

tests/unit/test_package.py
kokosing: it was by purpose, `l` caused flake to fail and I didn't want to break this line ;) Now, once I line length limit is extended I will fix that ;)

prestoadmin/package.py
kokosing: Well, not always. If _rpm_install fails the exception should be raised, unless it is called `with quite()` then this will be printed.

Or I don't understand something. 

Anyway, I will revert this check.

docs/presto-admin-commands.rst
kokosing: done

prestoadmin/package.py
kokosing: done

prestoadmin/package.py
kokosing: fixed

prestoadmin/package.py
kokosing: fixed

prestoadmin/package.py
rschlussel: sudo() in _rpm_install won't fail even if the command you ran has a non-zero exit code.  It only fails if there's a connection issue or something trying to run the command such that the function call itself has an issue.  But the thing that sudo() returns has a field to say whether the command succeeded or failed. That's why we needed to check the return code.

kokosing:
@rschlussel Comments applied.

tests/unit/test_package.py
```diff
@@ -51,39 +51,22 @@ def test_rpm_install_nodeps(self, mock_sudo):
...
-
-        mock_sudo.assert_any_call('rpm -qp --queryformat \'%{NAME}\' '
-                                  '/opt/prestoadmin/packages/test.rpm',
-                                  quiet=True)
-        mock_sudo.assert_any_call('rpm -i --nodeps '
-                                  '/opt/prestoadmin/packages/test.rpm')
-        mock_sudo.assert_any_call('rpm -e --nodeps '
-                                  '/opt/prestoadmin/packages/test.rpm')
-        self.assertEqual(3, mock_sudo.call_count)
+        mock_rpm_uninstal.assert_any_call('test_package_name')
```
rschlussel: name needs to be fixed here too

rschlussel:
looks good

ebd2:
FYI, `make precommit` runs flake, the docs build, the unit tests, etc. I.e. the things I usually forget to run one of if I run them by hand ;-)

PR #197, opened 2016-06-30
Display test execution time

kokosing:
Display test execution time

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@ envlist = py26, py27
 [testenv]
 setenv =
     PYTHONPATH = {toxinidir}:{toxinidir}/prestoadmin
-commands = nosetests {posargs}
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -4,7 +4,7 @@ envlist = py26, py27
 [testenv]
 setenv =
     PYTHONPATH = {toxinidir}:{toxinidir}/prestoadmin
-commands = nosetests {posargs}
+commands = nosetests --with-timer --timer-ok 60s --timer-warning 300s {posargs}
```
rschlussel: did you check that this is the right time scale for our product tests?  How many end up in each category?

rschlussel:
lgtm

tox.ini
kokosing: Currently a lot are yellow (lasts more that 60s) and some (1 per 10) are red (lasts more than 300s). 

I didn't want to align this numbers with what we have, but this is what I think we would like to have ;)

PR #198, opened 2016-06-30
Codestyle - let line length limit to be 120 chars

### LINKED ISSUE #191: Extend flake8 line length limit
Having line length limit to 80 characters is very inconvenient and causes a lot of lines to broken into multiple lines. 

I would extend it to 120, what do you think?

kokosing:
Codestyle - let line length limit to be 120 chars

Fixes #191

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,3 +2,5 @@
 universal = 0
 [nosetests]
 verbosity=3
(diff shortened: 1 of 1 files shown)


rschlussel:
lgtm

PR #199, opened 2016-07-06
Remove Presto version check in server commands

cawallin:
This check is no longer relevant, and fails whenever we change the
Presto version.

Additionally, modify version checking in the version range to allow
a 't' in the middle of version numbers (as in 0.148.t.0.1-SNAPSHOT).

### DIFF
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -17,7 +17,6 @@ Module for installing, monitoring and controlling presto server
 using presto-admin
 """
 import logging
-import re
 import sys
 
 from fabric.api import task, sudo, env, quiet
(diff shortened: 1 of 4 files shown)


cawallin:
@anusudarsan @rschlussel can you review this please?

cawallin:
Tested against an RPM with version 0.148.t.1-SNAPSHOT, and worked successfully.

rschlussel:
did you test that server status works correctly?

cawallin:
yeah, the tests pass and the output is as follows:
[root@master prestoadmin]# ./presto-admin server status
No mode selected. Defaulting to standalone. Select a mode using the command 'mode select <mode>'
Server Status:
	localhost(IP: 172.17.0.5, Roles: coordinator, worker): Running
	Node URI(http): http://172.17.0.5:8080
	Presto Version: 0.148-24-gf86354b
	Node status:    active
	Connectors:     system, tpch
[root@master prestoadmin]# rpm -qa | grep presto
presto-server-rpm-0.148.t.1.SNAPSHOT-1.x86_64

Which is expected since server status relies on the git tag rather than the RPM version.

rschlussel:
yup.  I was asking because we use the version checking to get the right sql for getting the nodes.

lgtm

prestoadmin/server.py
```diff
@@ -330,25 +327,6 @@ def check_presto_version():
         warn(not_installed_str)
         return not_installed_str
 
-    version = get_presto_version()
```
anusudarsan: can you update commit message, why this check is irrelevant now?

anusudarsan:
lgtm

PR #200, opened 2016-07-07
Enable travis to test all but product tests

kokosing:
Enable travis to test all but product tests


kokosing:
I am going to push this as it enable public and pre-commit test automation. 

However, any comments are still more than welcomed. They will be just applied as separate pull requests.

.travis.yml
```diff
@@ -0,0 +1,17 @@
+language: python
+python:
+  - "2.6"
+  - "2.7"
```
zachyee: Why are tests run against python 2.6 and 2.7 instead of just 2.7 as in the old travis script?

Issue #201, opened 2016-07-07
Remove usage of tox


kokosing:
tox requires that development workstation is equipped with two python installations (2.6 and 2.7). However it could be somehow difficult to achieve as some linux distribution no longer support usage of 2.6.

Notice that travis is able to test multiple version python.

cawallin:
If Travis supports it, we can definitely use that; it's rare that there are issues with 2.6/2.7 support. But we can't stop testing Python 2.6 because that's the ancient version of Python that the Hadoop appliance has.

ebd2:
I've been bit by 2.6/2.7 issues that tox has caught. I agree that we need to keep testing with 2.6, and tox is convenient for doing that locally.

PR #203, opened 2016-07-07
Fix travis - job timeout and test timeout

zachyee:
Created more jobs to fix the case where a single job would take more than 50 minutes to complete.

Moved creating the docker images for the product tests to before the test executes rather than during the test itself. This helps tests avoid their 10 minute limit for not outputting anything.

Added some logging that can be used in the future to figure out why tests are taking a long time (if this ever happens).

Works successfully with presto version 148, but not with the newest version. I can hardcode the URL where the tests download the presto rpm to fetch the 148 version rather than the newest version. This might be useful to test the release? (This PR fetches 149, not 148, so it will fail on travis)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,12 +6,50 @@ sudo: required
 services:
   - docker
 env:
-install: 
+  global:
+    - PA_TEST_ONLINE_INSTALLER=true
+    - PYTHONPATH=$(pwd)
+    - LONG_PRODUCT_TESTS="tests/product/test_server_install.py tests/product/test_status.py tests/product/test_collect.py tests/product/test_connectors.py tests/product/test_control.py tests/product/test_server_uninstall.py"
+  matrix:
+    - OTHER_TESTS=true
+    - SHORT_PRODUCT_TEST_GROUP=0
+    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN="tests/product/test_server_install.py"
+    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_status.py"
+    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_collect.py tests/product/test_server_uninstall.py"
+    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_connectors.py"
+    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_control.py"
+install:
   - pip install --upgrade pip==6.1.1
   - pip install -r requirements.txt
 before_script:
   - make docker-images
-script: 
-  - make clean lint dist docs 
-  - nosetests --with-timer --timer-ok 60s --timer-warning 300s -s tests.unit
-  - nosetests --with-timer --timer-ok 60s --timer-warning 300s -s tests.integration
+  - make presto-server-rpm.rpm
+script:
+  - |
+    if [ -v SHORT_PRODUCT_TEST_GROUP ]; then
+      ALL_PRODUCT_TESTS=$(find tests/product/ -name 'test_*py' | grep -v __init__ | xargs wc -l | sort -n | head -n -1 | awk '{print $2}' | tr '\n' ' ') &&
+      for LONG_PRODUCT_TEST in ${LONG_PRODUCT_TESTS[@]}; do
+        ALL_PRODUCT_TESTS=${ALL_PRODUCT_TESTS//$LONG_PRODUCT_TEST/};
+        if [ $? -ne 0 ]; then
+          exit 1
+        fi
+      done &&
+      SHORT_PRODUCT_TESTS=$(echo $ALL_PRODUCT_TESTS | tr ' ' '\n' | awk "NR % 1 == $SHORT_PRODUCT_TEST_GROUP" | tr '\n' ' ') &&
+      make test-images &&
+      nosetests --with-timer --timer-ok 60s --timer-warning 300s -a '!quarantine,!offline_installer' ${SHORT_PRODUCT_TESTS};
+    elif [ -v LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN ]; then
+      export IMAGE_NAMES="standalone_presto_admin" &&
+      make test-images &&
+      nosetests --with-timer --timer-ok 60s --timer-warning 300s -a '!quarantine,!offline_installer' ${LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN};
+    elif [ -v LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO ]; then
+      export IMAGE_NAMES="standalone_presto standalone_presto_admin" &&
+      make test-images &&
+      nosetests --with-timer --timer-ok 60s --timer-warning 300s -a '!quarantine,!offline_installer' ${LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO};
+    elif [ -v OTHER_TESTS ]; then
+      make clean lint dist docs test-rpm &&
+      nosetests -s tests.unit &&
+      nosetests -s tests.integration;
+    else
+      echo "Unknown test" &&
+      exit 1;
+    fi
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .PHONY: clean-all clean clean-eggs clean-build clean-pyc clean-test-containers clean-test \
-	clean-docs lint smoke test test-all test-rpm coverage docs open-docs release release-builds \
-	dist dist-online dist-offline wheel install precommit
+	clean-docs lint smoke test test-all test-images test-rpm docker-images coverage docs \
+	open-docs release release-builds dist dist-online dist-offline wheel install precommit
 
 help:
 	@echo "precommit - run \`quick' tests and tasks that should pass or succeed prior to pushing"
@@ -16,7 +16,9 @@ help:
 	@echo "smoke - run tests annotated with attr smoke using nosetests"
 	@echo "test - run tests quickly with Python 2.6 and 2.7"
 	@echo "test-all - run tests on every Python version with tox. Specify TEST_SUITE env variable to run only a given suite."
+	@echo "test-images - create product test image(s). Specify IMAGE_NAMES env variable to create only certain images."
 	@echo "test-rpm - run tests for the RPM package"
+	@echo "docker-images - pull docker image(s). Specify DOCKER_IMAGE_NAME env variable for specific image."
 	@echo "coverage - check code coverage quickly with the default Python"
 	@echo "docs - generate Sphinx HTML documentation, including API docs"
 	@echo "open-docs - open the root document (index.html) using xdg-open"
@@ -84,11 +86,15 @@ test-all: clean-test docker-images
 	tox -- -s tests.integration
 	tox -e py26 -- -s ${TEST_SUITE} -a '!quarantine'
 
+IMAGE_NAMES?="all"
+
+test-images: docker-images presto-server-rpm.rpm
+	python tests/product/image_builder.py ${IMAGE_NAMES}
 
 docker-images:
 	docker pull teradatalabs/centos6-ssh-oj8
 
-test-rpm: clean-test
+test-rpm: clean-test test-images
 	tox -e py26 -- -s tests.rpm -a '!quarantine'
 
 coverage:
--- a/Makefile
+++ b/Makefile
@@ -72,6 +72,9 @@ clean-docs:
 lint:
 	flake8 prestoadmin packaging tests
 
+presto-server-rpm.rpm:
+	wget 'https://repository.sonatype.org/service/local/artifact/maven/content?r=central-proxy&g=com.facebook.presto&a=presto-server-rpm&e=rpm&v=RELEASE' -O $@
+
 smoke: clean-test docker-images
 	tox -e py26 -- -a smoketest,'!quarantine'
 
--- a/tests/product/base_product_case.py
+++ b/tests/product/base_product_case.py
@@ -26,15 +26,12 @@ from nose.tools import nottest
 
 from prestoadmin.util import constants
 from tests.base_test_case import BaseTestCase
-from tests.configurable_cluster import ConfigurableCluster
-from tests.docker_cluster import DockerCluster, DockerClusterException
+from tests.docker_cluster import DockerCluster
 
-from tests.product.mode_installers import StandaloneModeInstaller, \
-    YarnSliderModeInstaller
-from tests.product.prestoadmin_installer import PrestoadminInstaller
+from tests.configurable_cluster import ConfigurableCluster
+from tests.product.cluster_types import cluster_types
 from tests.product.standalone.presto_installer import StandalonePrestoInstaller
-from tests.product.topology_installer import TopologyInstaller
-from tests.product.yarn_slider.slider_installer import SliderInstaller
+
 
 PRESTO_VERSION = r'.+'
 RETRY_TIMEOUT = 120
@@ -42,29 +39,6 @@ RETRY_INTERVAL = 5
 
 
 class BaseProductTestCase(BaseTestCase):
-    STANDALONE_BARE_CLUSTER = 'bare'
-    BARE_CLUSTER = 'bare'
-    PA_ONLY_CLUSTER = 'pa_only_standalone'
-    STANDALONE_PRESTO_CLUSTER = 'presto'
-
-    PA_ONLY_YS_CLUSTER = 'pa_only_ys'
-    PA_SLIDER_CLUSTER = 'pa_slider'
-
-    _cluster_types = {
-        BARE_CLUSTER: [],
-        PA_ONLY_CLUSTER: [PrestoadminInstaller,
-                          StandaloneModeInstaller],
-        STANDALONE_PRESTO_CLUSTER: [PrestoadminInstaller,
-                                    StandaloneModeInstaller,
-                                    TopologyInstaller,
-                                    StandalonePrestoInstaller],
-        PA_ONLY_YS_CLUSTER: [PrestoadminInstaller,
-                             YarnSliderModeInstaller],
-        PA_SLIDER_CLUSTER: [PrestoadminInstaller,
-                            YarnSliderModeInstaller,
-                            SliderInstaller]
-    }
-
     default_workers_config_ = """coordinator=false
 discovery.uri=http://master:8080
 http-server.http.port=8080
@@ -134,19 +108,11 @@ query.max-memory=50GB\n"""
         self.cluster = None
         self.default_keywords = {}
 
-    def _run_installers(self, installers):
-        cluster = self.cluster
-        for installer in installers:
-            dependencies = installer.get_dependencies()
-
-            for dependency in dependencies:
-                dependency.assert_installed(self)
-
-            installer_instance = installer(self)
-            installer_instance.install()
-
-            self.default_keywords.update(installer_instance.get_keywords())
-            cluster.postinstall(installer)
+    def tearDown(self):
+        self.restore_stdout_stderr_keep_open()
+        if self.cluster:
+            self.cluster.tear_down()
+        super(BaseProductTestCase, self).tearDown()
 
     def _apply_post_install_hooks(self, installers):
         for installer in installers:
@@ -158,12 +124,7 @@ query.max-memory=50GB\n"""
             self.default_keywords.update(installer_instance.get_keywords())
 
     def setup_cluster(self, bare_image_provider, cluster_type):
-        try:
-            installers = self._cluster_types[cluster_type]
-        except KeyError:
-            self.fail(
-                '%s is not a valid cluster type. Valid cluster types are %s' %
-                (cluster_type, ', '.join(self._cluster_types.keys())))
+        installers = cluster_types[cluster_type]
 
         config_filename = ConfigurableCluster.check_for_cluster_config()
 
@@ -172,42 +133,22 @@ query.max-memory=50GB\n"""
                 config_filename, self,
                 StandalonePrestoInstaller.assert_installed)
         else:
-            try:
-                self.cluster, bare_cluster = DockerCluster.start_cluster(
-                    bare_image_provider, cluster_type)
+            self.cluster, bare_cluster = DockerCluster.start_cluster(
+                bare_image_provider, cluster_type)
 
-                # If we've found images and started a non-bare cluster, the
-                # containers have already had the installers applied to them.
-                # We do need to get the test environment in sync with the
-                # containers by calling the following two functions.
-                #
-                # Once that's done, the cluster and test environment is in the
-                # same state it would be as if we'd called _run_installers on
-                # a bare cluster, and we can return.
-                #
-                # We do this to save the cost of running the installers on the
-                # docker containers every time we run a test. In practice,
-                # that turns out to be a fairly expensive thing to do.
-                if not bare_cluster:
-                    self._apply_post_install_hooks(installers)
-                    self._update_replacement_keywords(installers)
-                    return
-            except DockerClusterException as e:
-                self.fail(e.msg)
-
-        # If we got a bare cluster back, we need to run the installers on it.
-        # applying the post-install hooks and updating the replacement
-        # keywords is handled internally in _run_installers.
-        self._run_installers(installers)
-
-        if isinstance(self.cluster, DockerCluster):
-            self.cluster.commit_images(bare_image_provider, cluster_type)
-
-    def tearDown(self):
-        self.restore_stdout_stderr_keep_open()
-        if self.cluster:
-            self.cluster.tear_down()
-        super(BaseProductTestCase, self).tearDown()
+            # If we've found images and started a non-bare cluster, the
+            # containers have already had the installers applied to them.
+            # We do need to get the test environment in sync with the
+            # containers by calling the following two functions.
+            #
+            # We do this to save the cost of running the installers on the
+            # docker containers every time we run a test. In practice,
+            # that turns out to be a fairly expensive thing to do.
+            if not bare_cluster:
+                self._apply_post_install_hooks(installers)
+                self._update_replacement_keywords(installers)
+            else:
+                raise RuntimeError("Docker images have not been created")
 
     def dump_and_cp_topology(self, topology, cluster=None):
         if not cluster:
--- a/tests/product/image_builder.py
+++ b/tests/product/image_builder.py
@@ -1,0 +1,113 @@
+from tests.docker_cluster import DockerCluster
+from tests.hdp_bare_image_provider import HdpBareImageProvider
+from tests.no_hadoop_bare_image_provider import NoHadoopBareImageProvider
+
+from tests.product.base_product_case import BaseProductTestCase
+from tests.product.cluster_types import STANDALONE_BARE_CLUSTER, STANDALONE_PA_CLUSTER, \
+    STANDALONE_PRESTO_CLUSTER, YARN_SLIDER_PA_CLUSTER, cluster_types
+
+import argparse
+
+
+class ImageBuilder:
+    def __init__(self, testcase):
+        self.testcase = testcase
+        self.testcase.default_keywords = {}
+        self.testcase.cluster = None
+
+    def _run_installers(self, installers):
+        cluster = self.testcase.cluster
+        for installer in installers:
+            dependencies = installer.get_dependencies()
+
+            for dependency in dependencies:
+                dependency.assert_installed(self.testcase)
+
+            installer_instance = installer(self.testcase)
+            installer_instance.install()
+
+            self.testcase.default_keywords.update(installer_instance.get_keywords())
+            cluster.postinstall(installer)
+
+    def _setup_image(self, bare_image_provider, cluster_type):
+        installers = cluster_types[cluster_type]
+
+        self.testcase.cluster, bare_cluster = DockerCluster.start_cluster(
+            bare_image_provider, cluster_type)
+
+        # If we got a bare cluster back, we need to run the installers on it.
+        # applying the post-install hooks and updating the replacement
+        # keywords is handled internally in _run_installers.
+        #
+        # If we got a non-bare cluster back, that means the image already exists
+        # and we created the cluster using that image.
+        if bare_cluster:
+            self._run_installers(installers)
+
+            if isinstance(self.testcase.cluster, DockerCluster):
+                self.testcase.cluster.commit_images(bare_image_provider, cluster_type)
+
+        self.testcase.cluster.tear_down()
+
+    def _setup_image_with_no_hadoop_provider(self, cluster_type):
+        self._setup_image(NoHadoopBareImageProvider(),
+                          cluster_type)
+
+    def setup_standalone_presto_images(self):
+        cluster_type = STANDALONE_PRESTO_CLUSTER
+        self._setup_image_with_no_hadoop_provider(cluster_type)
+
+    def setup_standalone_presto_admin_images(self):
+        cluster_type = STANDALONE_PA_CLUSTER
+        self._setup_image_with_no_hadoop_provider(cluster_type)
+
+    def setup_standalone_bare_images(self):
+        cluster_type = STANDALONE_BARE_CLUSTER
+        self._setup_image_with_no_hadoop_provider(cluster_type)
+
+    def setup_yarn_slider_presto_admin_images(self):
+        cluster_type = YARN_SLIDER_PA_CLUSTER
+        self._setup_image_with_no_hadoop_provider(cluster_type)
+
+    def _setup_cluster_with_hdp_image_provider(self, cluster_type):
+        self._setup_image(HdpBareImageProvider(),
+                          cluster_type)
+
+    def setup_yarn_slider_presto_admin_and_slider_images(self):
+        cluster_type = self.testcase.YARN_SLIDER_PA_AND_SLIDER_CLUSTER
+        self._setup_cluster_with_hdp_image_provider(cluster_type)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("image_type", metavar="image_type", type=str, nargs="+",
+                        choices=["standalone_presto", "standalone_presto_admin",
+                                 "standalone_bare", "yarn_slider_presto_admin",
+                                 "all"],
+                        help="Specify the type of image to create. The available choices are: "
+                             "standalone_presto, standalone_presto_admin, standalone_bare, "
+                             "yarn_slider_presto_admin, all")
+
+    # ImageBuilder needs an input testcase with access to unittest assertions
+    # so the installers can check their resulting installations as well as some
+    # product test helper functions.
+    # This supplies a dummy testcase. BaseProductTestCase inherits from
+    # unittest. A unittest instance can be successfully created if the name
+    # of an existing method of the class is passed into the constructor.
+    dummy_testcase = BaseProductTestCase('__init__')
+    image_builder = ImageBuilder(dummy_testcase)
+
+    args = parser.parse_args()
+    if "all" in args.image_type:
+        image_builder.setup_standalone_presto_images()
+        image_builder.setup_standalone_presto_admin_images()
+        image_builder.setup_standalone_bare_images()
+        image_builder.setup_yarn_slider_presto_admin_images()
+    else:
+        if "standalone_presto" in args.image_type:
+            image_builder.setup_standalone_presto_images()
+        if "standalone_presto_admin" in args.image_type:
+            image_builder.setup_standalone_presto_admin_images()
+        if "standalone_bare" in args.image_type:
+            image_builder.setup_standalone_bare_images()
+        if "yarn_slider_presto_admin" in args.image_type:
+            image_builder.setup_yarn_slider_presto_admin_images()
--- a/tests/product/standalone/presto_installer.py
+++ b/tests/product/standalone/presto_installer.py
@@ -18,7 +18,6 @@ Module for installing presto on a cluster.
 
 import fnmatch
 import os
-import urllib
 
 import prestoadmin
 
@@ -102,26 +101,9 @@ class StandalonePrestoInstaller(BaseInstaller):
             cluster = self.testcase.cluster
 
         rpm_path = os.path.join(self.rpm_dir, self.rpm_name)
-        try:
-            cluster.copy_to_host(rpm_path, cluster.master)
-            self._check_if_corrupted_rpm(self.rpm_name, cluster)
-        except OSError:
-            print 'Downloading RPM again'
-            # try to download the RPM again if it's corrupt (but only once)
-            StandalonePrestoInstaller._download_rpm()
-            cluster.copy_to_host(rpm_path, cluster.master)
-            self._check_if_corrupted_rpm(self.rpm_name, cluster)
+        cluster.copy_to_host(rpm_path, cluster.master)
+        self._check_if_corrupted_rpm(self.rpm_name, cluster)
         return self.rpm_name
-
-    @staticmethod
-    def _download_rpm():
-        rpm_filename = 'presto-server-rpm.rpm'
-        rpm_path = os.path.join(prestoadmin.main_dir,
-                                rpm_filename)
-        urllib.urlretrieve(
-            'http://search.maven.org/remotecontent?filepath=com/facebook/presto/'
-            'presto-server-rpm/0.148/presto-server-rpm-0.148.rpm', rpm_path)
-        return rpm_filename
 
     @staticmethod
     def _detect_presto_rpm():
@@ -136,11 +118,7 @@ class StandalonePrestoInstaller(BaseInstaller):
             # are multiple RPMs, the last one is probably the latest
             rpm_name = sorted(rpm_names)[-1]
         else:
-            try:
-                rpm_name = StandalonePrestoInstaller._download_rpm()
-            except:
-                # retry once
-                rpm_name = StandalonePrestoInstaller._download_rpm()
+            raise OSError(1, 'Presto RPM not detected.')
 
         return prestoadmin.main_dir, rpm_name
 
--- a/tests/product/standalone/presto_installer.py
+++ b/tests/product/standalone/presto_installer.py
@@ -119,9 +119,8 @@ class StandalonePrestoInstaller(BaseInstaller):
         rpm_path = os.path.join(prestoadmin.main_dir,
                                 rpm_filename)
         urllib.urlretrieve(
-            'https://repository.sonatype.org/service/local/artifact/maven'
-            '/content?r=central-proxy&g=com.facebook.presto'
-            '&a=presto-server-rpm&e=rpm&v=RELEASE', rpm_path)
+            'http://search.maven.org/remotecontent?filepath=com/facebook/presto/'
+            'presto-server-rpm/0.148/presto-server-rpm-0.148.rpm', rpm_path)
         return rpm_filename
 
     @staticmethod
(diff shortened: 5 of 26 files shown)


.travis.yml
```diff
@@ -0,0 +1,55 @@
...
+sudo: required
+services:
+  - docker
+env:
+  global:
+    - PA_TEST_ONLINE_INSTALLER=true
+    - PYTHONPATH=$(pwd)
+  matrix:
+    - OTHER_TESTS=true
+    - PRODUCT_TEST_GROUP_SA_BARE="tests/product/test_installer.py tests/product/standalone/test_installation.py"
```
petroav: What does SA stand for?

.travis.yml
```diff
@@ -0,0 +1,55 @@
...
+    - PRODUCT_TEST_GROUP_SA_PRESTO="tests/product/test_authentication.py tests/product/test_configuration.py tests/product/test_server_upgrade.py"
+    - PRODUCT_TEST_GROUP_SA_PA="tests/product/test_error_handling.py tests/product/test_package_install.py tests/product/test_plugin.py tests/product/test_script.py tests/product/test_topology.py"
+    - PRODUCT_TEST_GROUP_SA_PA="tests/product/test_status.py"
+    - PRODUCT_TEST_GROUP_SA_PA="tests/product/test_server_install.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_collect.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_connectors.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_control.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_server_uninstall.py"
+    - PRODUCT_TEST_GROUP_YS_PA="tests/product/yarn_slider/test_slider_installation.py"
+    #-QUARANTINED="tests/product/yarn_slider/test_server_install.py tests/product/test_create_images.py"
```
petroav: Why is this commented out?

.travis.yml
zachyee: standalone

.travis.yml
```diff
@@ -0,0 +1,55 @@
...
+      make test-images
+      tox -e py26 -- -a '!quarantine,!offline_installer' ${PRODUCT_TEST_GROUP_SA_PA_PRESTO}
+    elif [ -v PRODUCT_TEST_GROUP_YS_PA ]; then
+      export IMAGE_NAMES="yarn_slider_pa_only"
+      make test-images
+      tox -e py26 -- -a '!quarantine,!offline_installer' ${PRODUCT_TEST_GROUP_YS_PA}
+    elif [ -v OTHER_TESTS ]; then
+      make clean lint docs test test-rpm
+    else
+      echo "Unknown test"
```
petroav: I think this will succeed the build. We want it to fail. Can you change it so that the build becomes red?

.travis.yml
zachyee: I can uncomment it. They're the quarantined tests, so they don't end up running but would trigger another job for travis to do.

Makefile
```diff
@@ -84,6 +86,10 @@ test-all: clean-test docker-images
 	tox -- -s tests.integration
 	tox -e py26 -- -s ${TEST_SUITE} -a '!quarantine'
 
+IMAGE_NAMES?="all"
```
petroav: What does the ? do?

tests/product/test_connectors.py
```diff
@@ -297,7 +300,7 @@ def test_connector_remove(self):
         for host in self.cluster.all_hosts():
             self.assert_has_default_connector(host)
 
-        missing_connector_message = """[Errno 1] 
+        missing_connector_message = """[Errno 1]
```
petroav: Did you intend to remove this single space character? I think this test does a string match and will fail if the error doesn't exactly correspond to what is expected.

Makefile
zachyee: if the variable hasn't been exported, it will default to setting it to "all". Otherwise it uses the exported value

tests/product/test_create_images.py
```diff
@@ -0,0 +1,98 @@
...
+    def test_setup_yarn_slider_pa_only_images(self):
+        cluster_type = self.PA_ONLY_YS_CLUSTER
+        self._setup_cluster_with_no_hadoop_provider_assert_images(cluster_type)
+
+
+if __name__ == "__main__":
+    # Allow arguments to be quoted in a string or input as multiple arguments:
+    # python test_create_images.py "standalone_presto standalone_pa_only"
+    # python test_create_images.py standalone_presto standalone_pa_only
+    command_line_arguments = argv[1:]
```
petroav: You should use the argparse module for argument parsing https://docs.python.org/2/library/argparse.html.

tests/product/test_create_images.py
```diff
@@ -0,0 +1,98 @@
...
+from subprocess import check_output
+from sys import argv
+from tests.no_hadoop_bare_image_provider import NoHadoopBareImageProvider
+from tests.product.base_product_case import BaseProductTestCase
+
+import unittest
+
+
+@attr('quarantine')
+class TestSetupImages(BaseProductTestCase):
```
petroav: I see what you did here, it's pretty clever :) but a little hacky in my opinion. You call into `setup_cluster()` and piggyback on that infrastructure. However, the price you pay is that your methods need to be tests. I think we need to rip out `check_cluster()` from `BaseProductTestCase` so that this module can be completely standalone and doesn't need to run as a fake test.

.travis.yml
```diff
@@ -17,6 +17,13 @@ env:
     - PRODUCT_TEST_GROUP=5
     - PRODUCT_TEST_GROUP=6
     - PRODUCT_TEST_GROUP=7
+    - PRODUCT_TEST_GROUP=8
+    - PRODUCT_TEST_GROUP=9
+    - PRODUCT_TEST_GROUP=10
+    - PRODUCT_TEST_GROUP=11
+    - PRODUCT_TEST_GROUP=12
+    - PRODUCT_TEST_GROUP=13
+    - PRODUCT_TEST_GROUP=14
```
kokosing: travis not only fails because of timeouts. Travis has low amount of resources. There are tests which spin up a docker cluster which consists of 4 docker containers where each tries to start presto. In such case travis kills some processes as the memory usage is exceeding travis capacity. I think this is the most serious issue. 

Moreover, when a process is killed, then presto-admin waits until timeout to reports that presto-server didn't show up. Thus making tests last longer.

.travis.yml
```diff
@@ -9,21 +9,13 @@ env:
...
+    - PRODUCT_TEST_GROUP="tests/product/test_server_uninstall.py tests/product/topology_installer.py tests/product/test_server_upgrade.py tests/product/yarn_slider/slider_installer.py tests/product/test_authentication.py"
+    - PRODUCT_TEST_GROUP="tests/product/yarn_slider/test_server_install.py tests/product/test_control.py"
+    - PRODUCT_TEST_GROUP="tests/product/yarn_slider/test_slider_installation.py tests/product/test_connectors.py"
+    - PRODUCT_TEST_GROUP="tests/product/test_plugin.py tests/product/test_collect.py"
+    - PRODUCT_TEST_GROUP="tests/product/mode_installers.py tests/product/test_status.py tests/product/test_error_handling.py tests/product/standalone/test_installation.py"
+    - PRODUCT_TEST_GROUP="tests/product/yarn_slider/pa_slider_config.py tests/product/test_package_install.py tests/product/test_topology.py tests/product/test_configuration.py"
```
kokosing: This not the way to go I think. It would require each developer to remember to modify this groups whenever new test is added or even modified.

tests/product/timing_test_decorator.py
```diff
@@ -0,0 +1,65 @@
...
+    def name_wrapper(some_function):
+        def time_wrapper(*args, **kwargs):
+            logger = _initialize_module_logger(module_name)
+            function_name = some_function.__name__
+            function_is_a_test = _name_starts_with_test(function_name)
+            if function_is_a_test:
+                logger.info("%s started", function_name)
+            start_time = time()
+
+            return_value = some_function(*args, **kwargs)
```
kokosing: what if function raises an exception?

tests/product/timing_test_decorator.py
```diff
@@ -0,0 +1,65 @@
+import logging
```
kokosing: How does it relate to `https://github.com/prestodb/presto-admin/commit/af804f7384e3ab5af4efe667398fff2e18060c62`? Can we revert that commit?

Makefile
```diff
@@ -59,6 +59,7 @@ clean-test:
 	rm -f .coverage
 	rm -fr htmlcov/
 	rm -fr tmp
+	find ./tests/product -name "*.log" -type f -delete
```
kokosing: why log files are created there? no temporary files should be created in source directory. Maybe instead of hiding the issue, please fix the location where log files are created.

tests/product/test_create_images.py
```diff
@@ -0,0 +1,91 @@
...
+            assert expected_image in docker_images
+
+    def _setup_cluster_with_no_hadoop_provider_assert_images(self, cluster_type):
+        self.setup_cluster(NoHadoopBareImageProvider(),
+                           cluster_type)
+        self._assert_expected_images_created(cluster_type)
+
+    def test_setup_standalone_presto_images(self):
+        cluster_type = self.STANDALONE_PRESTO_CLUSTER
+        self._setup_cluster_with_no_hadoop_provider_assert_images(cluster_type)
```
kokosing: I don't understand why do we use python to build docker images. I would prefer to have make to call docker directly, this would be much easier to test and extend. 

I tried this approach here: https://github.com/prestodb/presto-admin/pull/195

tests/product/base_product_case.py
```diff
@@ -135,7 +135,6 @@ def setUp(self):
         self.maxDiff = None
         self.cluster = None
         self.default_keywords = {}
-        logging.disable(logging.NOTSET)
```
kokosing: squash that change with your previous commit which introduced that.

tests/product/base_product_case.py
```diff
@@ -18,10 +18,9 @@
 """
 
 import json
-import logging
 import os
 import re
-from time import sleep, clock
```
kokosing: squash that change with the commit which introduced the clock

kokosing:
Travis didn't pass ; /

.travis.yml
zachyee: This was to fix jobs going over 50 minutes and failing. I wasn't considering Travis memory usage as part of this PR.

tests/product/test_create_images.py
zachyee: committed a change to use this instead

tests/product/test_create_images.py
zachyee: i took the functions out of the test case and put them in the module. you can see if that's what you had in mind.

tests/product/timing_test_decorator.py
zachyee: well i don't catch them, so they will just bubble up the stack. is that a problem?

tests/product/timing_test_decorator.py
zachyee: i'm sorry; i don't see the problem. my idea was to use what you added and build off of that. for example, we run the tests and your change outputs how long the test takes. if it takes a long time, we can use the logger to time some of the functions that test uses and figure out where the problem is.

Makefile
zachyee: where should the log files be created?

tests/product/test_create_images.py
zachyee: the docker images aren't pulled from anywhere. the product tests pull down a bare image and build the images they need from there. i was building off of this framework. 

i think your approach in #195 does not help in the case?

zachyee:
haha yea i noted when i made the PR that travis still wouldn't work with the newest version of presto (which this branch uses). i created an experiment branch that uses 148 and those builds have worked.

.travis.yml
zachyee: this was an older commit to consolidate those product test groups so we didn't have so many jobs running. is the new way the groups are formatted still a problem? i grouped them by the docker images they need to generate before running the test.

tests/product/test_create_images.py
petroav: Calling Docker directly from Make and transferring all this logic to Dockerfiles is probably a good idea. However, I don't know if all the functionality in the current Python installers can be accurately transferred. This work would be a major refactor and I don't think we should do it as part of this effort. Let's file a ticket and plan for it at some point in the future.

tests/product/image_builder.py
```diff
@@ -0,0 +1,203 @@
...
+                        action="store_true")
+    parser.add_argument("--yarn_slider_pa",
+                        help="Create yarn-slider presto-admin only images",
+                        action="store_true")
+    parser.add_argument("--all",
+                        help="Create standalone presto, standalone presto-admin only,"
+                             "standalone bare, and yarn-slider presto-admin only images",
+                        action="store_true")
+
+    if len(argv) <= 1:
```
petroav: I think it is best to re-write the argument parsing to use a positional argument. That way you can more succinctly specify that you require at least one argument and you can specify what the available choices are. Here's what I propose:

```
if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("image_type", metavar="image_type", type=str, nargs="+",
                        choices=["standalone_presto", "standalone_pa", "standalone_bare",
                                 "yarn_slider_pa", "all"], help="Specify the type of "
                        "image to create. The available choices are: standalone_presto, "
                        "standalone_pa, standalone_bare, yarn_slider_pa, all")

    image_builder = ProductTestImageBuilder()

    args = parser.parse_args()
    if "all" in args.image_type:
        image_builder.test_setup_standalone_pa_only_images()
        image_builder.test_setup_standalone_pa_only_images()
        image_builder.test_setup_standalone_bare_images()
        image_builder.test_setup_yarn_slider_pa_only_images()
    else:
        if "standalone_presto" in args.image_type:
            image_builder.test_setup_standalone_presto_images()
        if "standalone_pa" in args.image_type:
            image_builder.test_setup_standalone_pa_only_images()
        if "standalone_bare" in args.image_type:
            image_builder.test_setup_standalone_bare_images()
        if "yarn_slider_pa" in args.image_type:
            image_builder.test_setup_yarn_slider_pa_only_images()
```

tests/product/image_builder.py
```diff
@@ -0,0 +1,203 @@
...
+    YarnSliderModeInstaller
+from tests.product.prestoadmin_installer import PrestoadminInstaller
+from tests.product.standalone.presto_installer import StandalonePrestoInstaller
+from tests.product.topology_installer import TopologyInstaller
+from tests.product.yarn_slider.slider_installer import SliderInstaller
+
+import argparse
+
+
+class ProductTestImageBuilder:
```
petroav: Can you rename this class to `ImageBuilder` so it matches the file name, or vice versa?

tests/product/image_builder.py
```diff
@@ -0,0 +1,203 @@
...
+                                    StandalonePrestoInstaller],
+        PA_ONLY_YS_CLUSTER: [PrestoadminInstaller,
+                             YarnSliderModeInstaller],
+        PA_SLIDER_CLUSTER: [PrestoadminInstaller,
+                            YarnSliderModeInstaller,
+                            SliderInstaller]
+    }
+
+    def __init__(self):
+        self.maxDiff = None
```
petroav: This appears unused. Can you double check and remove if that's the case?

tests/product/image_builder.py
```diff
@@ -0,0 +1,203 @@
...
+        self.cluster = None
+        self.default_keywords = {}
```
petroav: What is this field used for?

tests/product/image_builder.py
```diff
@@ -0,0 +1,203 @@
...
+                dependency.assert_installed(self)
+
+            installer_instance = installer(self)
+            installer_instance.install()
+
+            self.default_keywords.update(installer_instance.get_keywords())
+            cluster.postinstall(installer)
+
+    def _setup_cluster(self, bare_image_provider, cluster_type):
+        installers = None
```
petroav: I don't think this initialization is necessary.

tests/product/image_builder.py
```diff
@@ -0,0 +1,203 @@
...
+        expected_images = self._get_expected_image_names(cluster_type)
+        for expected_image in expected_images:
+            assert expected_image in docker_images
+
+    def _setup_cluster_with_no_hadoop_provider_assert_images(self, cluster_type):
+        self._setup_cluster(NoHadoopBareImageProvider(),
+                            cluster_type)
+        self._assert_expected_images_created(cluster_type)
+
+    def test_setup_standalone_presto_images(self):
```
petroav: The `test` prefix for all these methods is no longer needed since they're no longer tests.

tests/product/image_builder.py
zachyee: sure this seems a little cleaner

tests/product/image_builder.py
zachyee: yes, i'll rename the class to ImageBuilder

tests/product/image_builder.py
```diff
@@ -0,0 +1,203 @@
...
+            installer_instance.install()
+
+            self.default_keywords.update(installer_instance.get_keywords())
+            cluster.postinstall(installer)
+
+    def _setup_cluster(self, bare_image_provider, cluster_type):
+        installers = None
+        try:
+            installers = self._cluster_types[cluster_type]
+        except KeyError:
```
petroav: I would remove this `try except` for `KeyError` because this validation happens implicitly at the argument parsing level.

tests/product/image_builder.py
zachyee: _cluster_types is used at line 72.

    def _setup_cluster(self, bare_image_provider, cluster_type):
        installers = None
        try:
            installers = self._cluster_types[cluster_type]

petroav: ? I'm talking about the `maxDiff` field, not the `_cluster_types` field.

tests/product/image_builder.py
zachyee: for initialization, I can remove maxDiff.
I could also remove self.cluster since it should be set by setup_cluster()

default_keywords:

    def _update_replacement_keywords(self, installers):
        for installer in installers:
            installer_instance = installer(self)
            self.default_keywords.update(installer_instance.get_keywords())

    def _setup_cluster(self, bare_image_provider, cluster_type):
    ...
                if not bare_cluster:
                    self._apply_post_install_hooks(installers)
                    self._update_replacement_keywords(installers)

tests/product/image_builder.py
```diff
@@ -0,0 +1,203 @@
...
+                # a bare cluster, and we can return.
+                #
+                # We do this to save the cost of running the installers on the
+                # docker containers every time we run a test. In practice,
+                # that turns out to be a fairly expensive thing to do.
+                if not bare_cluster:
+                    self._apply_post_install_hooks(installers)
+                    self._update_replacement_keywords(installers)
+                    return
+            except DockerClusterException as e:
```
petroav: You can remove this `try except` here as well and let the `DockerClusterException` bubble out to the top and interrupt the program. The reason we were catching it before is that we wanted to make sure that we failed all the tests.

tests/product/image_builder.py
zachyee: what initialization?

tests/product/image_builder.py
petroav: As part of this refactor you should edit `BaseProductTest` to use this class to generate its cluster, otherwise we end up with a lot of code duplication. There's logic to create Docker images/clusters here and in `BaseProductTest`.

.travis.yml
zachyee: yup

tests/product/image_builder.py
petroav: > I could also remove self.cluster since it should be set by setup_cluster()

I don't if that's a good idea. I think you'll need a place to store the resulting cluster object and return it to whoever is using this class. See my comment about making `BaseProductTest` also use this.

For `default_keywords` I was looking for an explanation on what that field does. I looked at its usage but couldn't figure out what its purpose is on a quick glance.

tests/product/image_builder.py
petroav: Assigning `None` to the variable. You can delete this line.

tests/product/image_builder.py
zachyee: oh okay. yes i've gotten rid of `maxDiff`

Makefile
petroav: Very cool!

tests/product/image_builder.py
zachyee: Each installer type has some keywords that it can access. For instance, the presto installer will have:

`{
        'rpm': self.rpm_name,
        'rpm_basename': RPM_BASENAME
}`

.travis.yml
petroav: @kokosing what would be a better way to split up the tests?

tests/product/image_builder.py
petroav: I see. Thanks for the explanation!

tests/product/image_builder.py
zachyee: It had to be there because of the try-except block. I can get rid of because of the argparsing:
>I would remove this try except for KeyError because this validation happens implicitly at the argument parsing level.

tests/product/timing_test_decorator.py
```diff
@@ -0,0 +1,88 @@
...
+    Usage:
+         @log_function_time(__name__)
+         def my_function(params):
+
+    Returns: Prints the execution time of the decorated function to the
+    log file(s). It will always log as info the execution time to the
+    info log file. If the execution time exceeds 10 minutes, it will also
+    log it to the error log file.
+    """
+    def name_wrapper(some_function):
```
petroav: `some_function` is too colloquial, I would rename it to just `function`.

tests/product/timing_test_decorator.py
petroav: @kokosing this decorator is meant to be complementary to your commit. It can used to annotate some key functions and measure their runtime so that we can quickly identify which function causes a runtime increase when that happens.

However, I think we should get rid of the log files and instead print times on stderr or stdout. That would greatly simplify the code. Also we could not have any methods decorated by default and instead have a PyDoc section on the decorator telling future developers what methods should be annotated if they wish to debug timing problems.

petroav:
You should squash pretty much all of your commits. At the moment they're broken down at a level that is too granular. I think you should have two commits: one that refactors travis.yml as well as the refactor of `BaseProductTest` and another commit that introduces the timing decorator.

I would also like @ebd2 to take a look at this when he gets back on Monday.

.travis.yml
```diff
@@ -0,0 +1,56 @@
...
+    - PRODUCT_TEST_GROUP_SA_PA="tests/product/test_status.py"
+    - PRODUCT_TEST_GROUP_SA_PA="tests/product/test_server_install.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_collect.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_connectors.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_control.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_server_uninstall.py"
+    - PRODUCT_TEST_GROUP_YS_PA="tests/product/yarn_slider/test_slider_installation.py"
+    #-QUARANTINED="tests/product/yarn_slider/test_server_install.py"
+install:
+  - pip install --upgrade pip==6.1.1
```
ebd2: You and @kokosing are going to need to coordinate some on this. See #195

tests/product/image_builder.py
```diff
@@ -0,0 +1,203 @@
...
+
+        # If we got a bare cluster back, we need to run the installers on it.
+        # applying the post-install hooks and updating the replacement
+        # keywords is handled internally in _run_installers.
+        self._run_installers(installers)
+
+        if isinstance(self.cluster, DockerCluster):
+            self.cluster.commit_images(bare_image_provider, cluster_type)
+
+    def _get_expected_image_names(self, cluster_type):
```
ebd2: This is probably more cleanly done as having a dict of cluster_type to (master_image_name, slave_image_name) and returning dict[cluster_type].
You should be returning tuples instead of lists because tuples are immutable.
Also, you should probably extract the teradatalabs/pa_test portion of the name to a constant?

ebd2:
Just had a thought: Parallelizing the tests almost requires that we create the docker images prior to starting the tests, otherwise, it's highly likely that multiple tests will think they're the first one that needs a particular image. This will result in committing different containers to the same image names.

At best we burn a little extra disk space (multiple layers with presto installed, of which only one is :latest). At worst, we get intermittent errors owing to the race condition.

zachyee:
Made (hopefully all) the changes you guys asked for. Also cleaned up the commits.

Here are the builds (Builds 241-251): https://travis-ci.org/prestodb/presto-admin/builds
Some of the tests failed, but didn't error with timeout issues.

Again, those builds used presto version 0.148 instead of the newest version (which this branch uses), so this PR travis build will still fail.

tests/product/timing_test_decorator.py
```diff
@@ -0,0 +1,81 @@
...
+         By default, this is set to 'root' to use the root logger.
+         It is recommended that __name__ be passed into this function
+         for the most accurate logging.
+
+    Usage:
+         @log_function_time(__name__)
+         def my_function(params):
+
+    Returns: Prints the execution time of the decorated function to the
+    log file(s). It will always log as info the execution time to the
```
petroav: Update PyDoc since logs now go to the console.

tests/product/timing_test_decorator.py
```diff
@@ -0,0 +1,81 @@
...
+    info log file. If the execution time exceeds 10 minutes, it will also
+    log it to the error log file.
+    """
+    def name_wrapper(function):
+        def time_wrapper(*args, **kwargs):
+            # logging.disable(logging.NOTSET)
+            logger = _initialize_module_logger(module_name)
```
petroav: It looks like logging is initialized every time this decorator is used. It would be better if we found a way to initialize the logging module once only. This race condition, for lack of a better term, is why you have an `if` statement on line 19, to avoid adding a handler every time logging is setup. Unfortunately, I don't how to do that so I'm gonna have to think about it. @ebd2 do you have any ideas?

tests/product/timing_test_decorator.py
```diff
@@ -0,0 +1,81 @@
...
+         def my_function(params):
+
+    Returns: Prints the execution time of the decorated function to the
+    log file(s). It will always log as info the execution time to the
+    info log file. If the execution time exceeds 10 minutes, it will also
+    log it to the error log file.
+    """
+    def name_wrapper(function):
+        def time_wrapper(*args, **kwargs):
+            # logging.disable(logging.NOTSET)
```
petroav: It's hacky to have people uncomment this whenever they want to use the decorator. We should enable logging automatically whenever the decorator is used, so this should probably go in the `_initialize_module_logger` method.

tests/product/timing_test_decorator.py
```diff
@@ -0,0 +1,81 @@
...
+            logger = _initialize_module_logger(module_name)
+            function_name = function.__name__
+            function_is_a_test = _name_starts_with_test(function_name)
```
petroav: What's the purpose of this? For test methods we explicitly print a statement saying that the test started but for other methods we don't? I don't see much value in making a distinction.

tests/product/timing_test_decorator.py
```diff
@@ -0,0 +1,81 @@
...
+            if function_is_a_test:
+                logger.info("%s started", function_name)
+            start_time = time()
+            logging.disable(logging.CRITICAL)
```
petroav: Why are you explicitly disabling `CRITICAL`?

tests/product/timing_test_decorator.py
```diff
@@ -0,0 +1,81 @@
...
+
+            return_value = function(*args, **kwargs)
+
+            # logging.disable(logging.NOTSET)
+            elapsed_time = time() - start_time
+            travis_output_time_limit = 600
+            message_level = logging.ERROR if elapsed_time >= travis_output_time_limit \
+                else logging.INFO
+            if function_is_a_test:
```
petroav: Yea, I don't see much value in making this distinction.

.travis.yml
```diff
@@ -6,12 +6,54 @@ sudo: required
...
+    - PRODUCT_TEST_GROUP_SA_PRESTO="tests/product/test_authentication.py tests/product/test_configuration.py tests/product/test_server_upgrade.py"
+    - PRODUCT_TEST_GROUP_SA_PA="tests/product/test_error_handling.py tests/product/test_package_install.py tests/product/test_plugin.py tests/product/test_script.py tests/product/test_topology.py"
+    - PRODUCT_TEST_GROUP_SA_PA="tests/product/test_server_install.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_status.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_collect.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_connectors.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_control.py"
+    - PRODUCT_TEST_GROUP_SA_PA_PRESTO="tests/product/test_server_uninstall.py"
+    - PRODUCT_TEST_GROUP_YS_PA="tests/product/yarn_slider/test_slider_installation.py"
+    #-QUARANTINED="tests/product/yarn_slider/test_server_install.py"
```
petroav: I think we should remove this. In general we shouldn't check in code that has been commented out.

tests/product/timing_test_decorator.py
zachyee: `logging.disable(logging.CRITICAL` disables logging for all messages with level <= logging.CRITICAL, so it pretty much disables all logging.

The tests disable logging, so I thought the logger should respect that by only enabling logging whenever they needed to log something.

tests/product/timing_test_decorator.py
zachyee: I found the distinction helpful when I was working with a lot of tests, but since this decorator will likely be used infrequently and on single tests, I think you're right. I'll go ahead and get rid of the distinction.

zachyee:
Made some changes to the decorator.

Also redid .travis.yml to have short and long tests separated into different groups. Long tests are hardcoded into their own groups and can build images beforehand to avoid timeout issues. Short tests are partitioned into groups using number of lines as a heuristic. I think this is what you had in mind @kokosing ?

tests/product/image_builder.py
```diff
@@ -0,0 +1,171 @@
...
+            # same state it would be as if we'd called _run_installers on
+            # a bare cluster, and we can return.
+            #
+            # We do this to save the cost of running the installers on the
+            # docker containers every time we run a test. In practice,
+            # that turns out to be a fairly expensive thing to do.
+            if not bare_cluster:
+                self._apply_post_install_hooks(installers)
+                self._update_replacement_keywords(installers)
+                return self.caller.cluster
```
ebd2: I believe these need to be called every time we create a new cluster from existing images.

If that's no longer true, they shouldn't be called here, because we should always have a bare cluster when we get there.

tests/product/image_builder.py
```diff
@@ -0,0 +1,171 @@
...
+                                 "yarn_slider_pa", "yarn_slider_pa_slider", "all"],
+                        help="Specify the type of image to create. The available choices are: "
+                             "standalone_presto, standalone_pa, standalone_bare, "
+                             "yarn_slider_pa, yarn_slider_pa_slider, all")
+
+    # ImageBuilder needs an input testcase with access to unittest assertions
+    # so the installers can check their resulting installations as well as some
+    # product test helper functions.
+    # This supplies a dummy testcase
+    dummy_testcase = BaseProductTestCase('__init__')
```
ebd2: What is this witchcraft?

Makefile
```diff
@@ -70,6 +72,9 @@ clean-docs:
 lint:
 	flake8 prestoadmin packaging tests
 
+get-presto-rpm:
```
ebd2: This should be in the form
```
path-to/presto-rpm:
        wget 'url' -o $@

File: tests/product/standalone/presto_installer.py [hunk 37]
```diff
@@ -116,12 +116,8 @@ def copy_presto_rpm_to_master(self, cluster=None):
     @staticmethod
     def _download_rpm():
         rpm_filename = 'presto-server-rpm.rpm'
-        rpm_path = os.path.join(prestoadmin.main_dir,
-                                rpm_filename)
-        urllib.urlretrieve(
-            'https://repository.sonatype.org/service/local/artifact/maven'
-            '/content?r=central-proxy&g=com.facebook.presto'
-            '&a=presto-server-rpm&e=rpm&v=RELEASE', rpm_path)
+        subprocess.check_call(['(cd ' + prestoadmin.main_dir +
```
ebd2: Why not just make this a dependency in the Makefile and error out here.
Putting this here hides the dependency relationship.
If we've gotten here without the dependency being satisfied, we should fail.

Makefile
ebd2: Oh, and it would also need to not be phony.

tests/product/image_builder.py
ebd2: Just got to the test cases. I don't think the testcases should be calling into image_builder. By the time the tests start running, it should be a catastrophic error if the images don't exist. As-is, if they don't exist, we fall back to the previous behavior.

Makefile
zachyee: Oh cool, I didn't know this was possible.

tests/product/image_builder.py
zachyee: It's the dummy product test case since the installers need one to invoke assertions like assertIn() and some of the product test helper functions like upload_topology()

tests/product/image_builder.py
zachyee: If the test cases don't call into image_builder, they won't have their cluster attribute properly set, right?

zachyee: I don't fully understand what happens in the code. I tore this straight out of BaseProductTestCase. Are you saying that:
>            if not bare_cluster:
>                 self._apply_post_install_hooks(installers)
>                 self._update_replacement_keywords(installers)
>                 return self.caller.cluster

Will never get run because we shouldn't ever have a bare cluster? If so, why is that the case?

.travis.yml
kokosing: it is not clear that SA means that, what about YS? Please do not use shortcuts.

.travis.yml
```diff
@@ -6,12 +6,54 @@ sudo: required
...
+      make test-images
+      tox -e py26 -- -a '!quarantine,!offline_installer' ${PRODUCT_TEST_GROUP_SA_PA_PRESTO}
+    elif [ -v PRODUCT_TEST_GROUP_YS_PA ]; then
+      export IMAGE_NAMES="yarn_slider_pa"
+      make test-images
+      tox -e py26 -- -a '!quarantine,!offline_installer' ${PRODUCT_TEST_GROUP_YS_PA}
+    elif [ -v OTHER_TESTS ]; then
+      make clean lint dist docs test-rpm
+      nosetests --with-timer --timer-ok 60s --timer-warning 300s -s tests.unit
+      nosetests --with-timer --timer-ok 60s --timer-warning 300s -s tests.integration
```
kokosing: I think there is no need to set timers for unit and integration tests, they are really short.

kokosing:
20 jobs travis sounds like an overkill, note that you are running `tox -e p26` on machines which requested python 2.7

tests/product/image_builder.py
ebd2: Building the images and creating clusters from those images should be separate. The test code shouldn't be calling into code that will fall back to creating the images if they haven't already been created. The code that handles actually bringing up a cluster should probably be in BaseProductCase.

tests/product/timing_test_decorator.py
```diff
@@ -0,0 +1,60 @@
+import logging
+import sys
+
+from time import time
+
+
+def _initialize_module_logger(module_name):
```
petroav: So I suggest we remove this method completely and make this piece of code execute at the module level. In the functions where you refer to the `logger` variable make sure you include the line `global logger`. Also, I suggest we hardcode the module to "root". I think that's fine because the reason why you would want to have different loggers for every module is to be able to adjust the logging level and logging destination at a more granular level whereas we just print everything at INFO to the console.

I did some experiments and determined that if the code is moved to be module level it will only get executed once.

tests/product/timing_test_decorator.py
```diff
@@ -0,0 +1,60 @@
...
+    logger.setLevel(logging.INFO)
+
+    message_format = '%(name)s - %(message)s'
+    formatter = logging.Formatter(message_format)
+
+    console_handler = logging.StreamHandler(sys.stdout)
+    console_handler.setLevel(logging.INFO)
+    console_handler.setFormatter(formatter)
+
+    if not len(logger.handlers):
```
petroav: This check can go away.

tests/product/timing_test_decorator.py
```diff
@@ -0,0 +1,60 @@
...
+        logger.addHandler(console_handler)
+    return logger
+
+
+def log_function_time(module_name="root"):
```
petroav: You can also remove this argument. Don't forget to update the PyDoc as well.

.travis.yml
petroav: Yea, maybe you can rename the groups to `STANDALONE_BARE_TESTS` and `YARN_SLIDER_TESTS`. Including the `PRODUCT_TEST_GROUP_*` prefix is a little too verbose and redundant.

.travis.yml
petroav: I agree.

tests/product/image_builder.py
petroav: I agree with Eric's point. In this piece of code if you do not get a bare cluster you should error out. The individual tests should call into BaseProductTest to setup a cluster and error out if they get a bare cluster. Hopefully this makes sense.

zachyee:
@kokosing good point. I think the tests are grouped as optimally as possible, given how long they take right now. I think the only solution is to rework the tests to run faster (like you said). It also doesn't help that the tests have to be run with Python 2.6, 2.7, which automatically doubles the number of jobs per build. There's no way to work around this is there?

@petroav is there a ticket filed somewhere for trying to make the tests fun quicker so we can have less jobs per build or should I do that? I also fixed up the logging in `timing_test_decorator.py`. You can double-check to make sure everything looks good there. 

@ebd2 i separated `image_builder.py` to just build images (`_setup_image()`) and gave the `BaseProductTestCase` a `setup_cluster()` function to use. Can you double-check that all the logic is correct for both of those functions?

petroav:
@zachyee nah, there's no ticket for speeding up the tests. Feel free to create one but don't work on it. If it takes 20 jobs to bring each run to under 50 minutes, or whatever the timeout is, then so be it. Have you tried with fewer jobs? Things might run faster now with the bigger machines so you should definitely see if you can optimize it.

tests/product/image_builder.py
```diff
@@ -0,0 +1,111 @@
...
+from tests.hdp_bare_image_provider import HdpBareImageProvider
+from tests.no_hadoop_bare_image_provider import NoHadoopBareImageProvider
+
+from tests.product.base_product_case import BaseProductTestCase
+
+import argparse
+
+
+class ImageBuilder:
+    def __init__(self, caller):
```
ebd2: The name caller suggests that it can be something pretty generic. I think we should do the following:

1. Let this take any testcase (and rename caller -> testcase)
2. Extract the cluster types into tests/product/cluster_types.py and import them here and in base_product_case.

That separates the two dependencies (testcase for assertions and cluster definitions) such that they can be satisfied independently.

tests/product/test_installer.py
```diff
@@ -32,7 +32,8 @@ class TestInstaller(BaseProductTestCase):
 
     def setUp(self):
         super(TestInstaller, self).setUp()
-        self.setup_cluster(NoHadoopBareImageProvider(), self.BARE_CLUSTER)
+        self.setup_cluster(NoHadoopBareImageProvider(),
```
ebd2: Do these need to be rewrapped? The line length limit is 120 now.

tests/product/base_product_case.py
```diff
@@ -44,25 +45,25 @@
...
                                     TopologyInstaller,
                                     StandalonePrestoInstaller],
-        PA_ONLY_YS_CLUSTER: [PrestoadminInstaller,
-                             YarnSliderModeInstaller],
-        PA_SLIDER_CLUSTER: [PrestoadminInstaller,
-                            YarnSliderModeInstaller,
-                            SliderInstaller]
+        YARN_SLIDER_PRESTO_ADMIN_CLUSTER: [PrestoadminInstaller,
+                                           YarnSliderModeInstaller],
+        YARN_SLIDER_PRESTO_ADMIN_AND_SLIDER_CLUSTER: [PrestoadminInstaller,
```
ebd2: I don't love the rename. This is adding a little more explicitness at the expense of reasonable length variable names than I'd like.

How about keeping PA abbreviated and leaving the expansion of YARN_SLIDER? Colloquially, presto admin is pa, and if you're writing code here you probably know that. Conversely, yarn and slider apply only to a subset of the people hacking the pa code.

ebd2:
Looks good modulo the small refactor and the naming nit.

kokosing:
@zachyee Please make sure that the problem from https://github.com/prestodb/presto/pull/5689 does not exists in your travis.yml script. 

Test grouping is ok. Thanks. 

Using tox and travis to provision different python version is redundant. We should choose one of these, I prefer travis. See https://github.com/prestodb/presto-admin/issues/201. You are using `tox -e py26...`  so even two version of python are requested only py26 is tested.

.travis.yml
```diff
@@ -6,12 +6,50 @@ sudo: required
...
 env:
-install: 
+  global:
+    - PA_TEST_ONLINE_INSTALLER=true
+    - PYTHONPATH=$(pwd)
+    - LONG_PRODUCT_TESTS="tests/product/test_server_install.py tests/product/test_status.py tests/product/test_collect.py tests/product/test_connectors.py tests/product/test_control.py tests/product/test_server_uninstall.py"
+  matrix:
+    - OTHER_TESTS=true
+    - SHORT_PRODUCT_TEST_GROUP=0
+    - LONG_PRODUCT_TEST_GROUP_STANDALONE_PRESTO_ADMIN="tests/product/test_server_install.py"
```
rschlussel: I'd get rid of the word standalone in all the test group names.  It's the only way to run presto-admin right now, and makes the name really long.  
(Sorry for chiming in so late, I've been using this patch to be able to run tests on travis, and noticed this.  Also--works great!  thanks!)

zachyee:
I combined a couple of job groups to get rid of 4 jobs (from 18 to 14). That's the best I can do until tests are faster or someone reaches out to Travis to increase the 50-minute timeout limit.

Did the cluster_type refactor and renaming.

@kokosing thanks for pointing that out. I ran into a similar problem when it wasn't erroring out from lint errors. I went ahead and `&&` commands together so it fails when the first command fails. It works fine for presto-admin because I made it one long if, elif, ... else script, whereas presto needs a bunch of separate scripts with single if statements. If this changes, we can use the fix introduced to presto, but I think this way is more readable. I also removed using `tox` and invoked `nosetests` directly. It will use whatever version is installed on the travis machine.

tests/product/standalone/presto_installer.py
```diff
@@ -137,11 +118,7 @@ def _detect_presto_rpm():
             # are multiple RPMs, the last one is probably the latest
             rpm_name = sorted(rpm_names)[-1]
         else:
-            try:
-                rpm_name = StandalonePrestoInstaller._download_rpm()
-            except:
-                # retry once
-                rpm_name = StandalonePrestoInstaller._download_rpm()
+            raise OSError(1, 'Presto RPM not detected.')
```
ebd2: nit: you have a testcase where this gets called in the constructor. Does it make sense to pass it in here and testcase.fail() if the RPM doesn't exist?

It might not because it isn't a real test case, if it's messier that way, I'm fine with it as-is.

ebd2:
LGTM!

rschlussel:
I wouldn't worry about decreasing the number of jobs further.  It's better for it not to need so many jobs, but having the whole thing complete in a reasonable amount of time is more important.

.travis.yml
petroav: +1

petroav:
Looks good

PR #204, opened 2016-07-07
Remove note about rpm version from emr docs

rschlussel:
We now build a single presto rpm that works for all versions of rpm

@anusudarsan


anusudarsan:
lgtm

rschlussel:
merged

PR #205, opened 2016-07-20
Get server status from coordinator by node.id

rschlussel:
Starting in Presto 0.149, the REST API on worker nodes is not
accessible, which means that we cannot test whether a server is started
by contacting that server directly.  Instead we get system.runtime.nodes
from the coordinator and check that the node.id for the current node is
listed.

### DIFF
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -18,6 +18,7 @@ using presto-admin
 """
 import logging
 import sys
+from tempfile import mkdtemp
 
 from fabric.api import task, sudo, env
 from fabric.context_managers import settings, hide
@@ -27,23 +28,21 @@ from fabric.tasks import execute
 from fabric.utils import warn, error, abort
 from retrying import retry
 
+import util.filesystem
 from prestoadmin import configure_cmds
 from prestoadmin import connector
 from prestoadmin import package
-from prestoadmin.util.version_util import VersionRange, VersionRangeList, \
-    split_version, strip_tag
 from prestoadmin.prestoclient import PrestoClient
 from prestoadmin.standalone.config import StandaloneConfig, \
     PRESTO_STANDALONE_USER_GROUP
-from prestoadmin.util.base_config import requires_config
 from prestoadmin.util import constants
-from prestoadmin.util.exception import ConfigFileNotFoundError
+from prestoadmin.util.base_config import requires_config
+from prestoadmin.util.exception import ConfigFileNotFoundError, ConfigurationError
 from prestoadmin.util.fabricapi import get_host_list, get_coordinator_role
 from prestoadmin.util.remote_config_util import lookup_port, \
-    lookup_server_log_file, lookup_launcher_log_file
-
-from tempfile import mkdtemp
-import util.filesystem
+    lookup_server_log_file, lookup_launcher_log_file, lookup_string_config
+from prestoadmin.util.version_util import VersionRange, VersionRangeList, \
+    split_version, strip_tag
 
 __all__ = ['install', 'uninstall', 'upgrade', 'start', 'stop', 'restart',
            'status']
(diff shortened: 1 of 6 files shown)


prestoadmin/server.py
```diff
@@ -27,23 +28,21 @@
 from fabric.utils import warn, error, abort
 from retrying import retry
 
+import util.filesystem
```
rschlussel: pycharm rearranged all the imports.  I wasn't sure if we always have our from imports after the regular imports, but the previous thing was definitely wrong since it was just a jumble.

prestoadmin/server.py
```diff
@@ -231,17 +230,18 @@ def service(control=None):
...
           'server does not respond.'
           % (env.host, (RETRY_TIMEOUT / 60)))
-    if check_server_status(client):
+    if check_server_status():
         print('Server started successfully on: ' + env.host)
     else:
-        error('Server failed to start on: ' + env.host +
-              '\nPlease check ' + lookup_server_log_file(env.host) + ' and ' +
-              lookup_launcher_log_file(env.host))
+        warn('Could not verify server status for: ' + env.host +
```
rschlussel: changed to warn since not being able to verify doesn't indicate something is wrong (could be something was wrong with the coordinator, e.g.).

prestoadmin/util/remote_config_util.py
```diff
@@ -86,7 +86,7 @@ def lookup_in_config(config_key, config_file, host):
                                warn_only=True, host=host)[host]
 
     if isinstance(config_value, Exception) or config_value.return_code == 2:
-        raise ConfigurationError('Configuration file %s does not exist on '
+        raise ConfigurationError('Could not access config file %s on '
```
rschlussel: The error also gets thrown for connection issues, so the previous error message was confusing

tests/product/test_control.py
```diff
@@ -134,17 +136,12 @@ def test_start_stop_restart_coordinator_down(self):
...
             self.cluster.internal_slaves[0])
 
     def test_start_stop_restart_worker_down(self):
-        installer = StandalonePrestoInstaller(self)
-        self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PA_CLUSTER)
-        topology = {"coordinator": "slave1",
-                    "workers": ["master", "slave2", "slave3"]}
-        self.upload_topology(topology=topology)
-        installer.install(coordinator='slave1')
+        self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PRESTO_CLUSTER)
```
rschlussel: The test was incorrect before.  It was testing coordinator down twice.

tests/product/test_control.py
```diff
@@ -230,40 +282,6 @@ def expected_down_node_output_size(self, expected_output):
         return self.len_down_node_error + len(
             '\n'.join(expected_output).splitlines())
 
-    def test_start_restart_config_file_error(self):
```
rschlussel: seemed more a test of presto than presto-admin

rschlussel:
@ebd2

prestoadmin/server.py
```diff
@@ -353,18 +353,33 @@ def check_server_status(client):
...
     time = 0
+
+    if len(get_coordinator_role()) < 1:
+        warn('No coordinator defined.  Cannot verify server status.')
+    client = PrestoClient(get_coordinator_role()[0], env.user)
+    node_id = lookup_string_config('node.id', os.path.join(constants.REMOTE_CONF_DIR, 'node.properties'), env.host)
     while time < RETRY_TIMEOUT:
-        result = client.execute_query(SYSTEM_RUNTIME_NODES)
-        if not result:
+        try:
```
ebd2: Can we use @retry for this like we do for wait_for_presto_user?

prestoadmin/server.py
```diff
@@ -353,18 +353,33 @@ def check_server_status(client):
...
         else:
-            break
-    return result
+            return True
+    return False
+
+
+def _is_in_rows(value, rows):
+    for row in rows:
+        if value in row:
```
ebd2: On the one hand, this doesn't depend on the current table layout, on the other, it's a little gross. I guess I'm OK with it though because of the layout independence.

tests/product/test_control.py
ebd2: Nice catch.

ebd2:
LGTM. If you want to punt the retry thing, feel free.

prestoadmin/server.py
rschlussel: i can do that in a separate pr.  I don't want to delay this getting in since without this fix we can't run any tests.

rschlussel:
travis failure looks like an unrelated intermittent issue where one of the servers didn't start up properly.  it passed in the 2.7 version of the same test

PR #206, opened 2016-07-20
Add ability to download rpm to server.install

zachyee:


docs/presto-admin-commands.rst
```diff
@@ -392,9 +392,14 @@ server install
 **************
 ::
 
-    presto-admin server install <local_path> [--nodeps]
+    presto-admin server install <rpm_specifier> [--nodeps]
 
-This command copies the presto-server rpm from ``local_path`` to all the nodes in the cluster, installs it, deploys the general presto configuration along with tpch connector configuration. The ``local_path`` should be accessible by ``presto-admin``.
+This command takes in a parameter ``rpm_specifier``, which must be a string describing the presto-server rpm to be used.
```
ebd2: Nice docs!

docs/presto-admin-commands.rst
```diff
@@ -392,9 +392,14 @@ server install
...
+The string can be a local path to a previously downloaded rpm. The local path should be accessible by ``presto-admin``.
+It may also be a version number (in the form X.Y or X.Y.Z when Z is applicable), a url, or 'release' (for the most recent release). In these cases, this command will attempt to find a local presto-server rpm that matches the requested version. If no match is found, it will attempt to download the rpm.
```
ebd2: how about 'latest' for the most recent version?

docs/presto-admin-commands.rst
```diff
@@ -411,6 +416,10 @@ Example
 ::
 
     sudo ./presto-admin server install /tmp/presto.rpm
+    sudo ./presto-admin server install 0.144.6
```
ebd2: How does this work with our 't'-branded Teradata versions?

prestoadmin/server.py
```diff
@@ -82,11 +86,191 @@ def new_sysnode_processor(node_info_rows):
...
+
+
+def _find_or_download_most_recent_presto_rpm():
+    newest_rpm_release = 'https://repository.sonatype.org/service/local/artifact/maven' \
+                         '/content?r=central-proxy&g=com.facebook.presto' \
+                         '&a=presto-server-rpm&e=rpm&v=RELEASE'
+    return _find_or_download_rpm(newest_rpm_release)
+
+
+def _check_if_valid_version(rpm_version):
```
ebd2: Document what this should and shouldn't match. Additionally: add unit tests if you haven't already.

ebd2: Also, is this really necessary? A valid version is whatever the maven repo will cough up, much as a valid email address is whatever you can deliver to.

prestoadmin/server.py
```diff
@@ -82,11 +86,191 @@ def new_sysnode_processor(node_info_rows):
...
+        return None
+
+
+def _get_download_file_name(url_response, version=None):
+    try:
+        headers = url_response.info()
+        content_disposition = headers['Content-Disposition']
+        values, params = cgi.parse_header(content_disposition)
+        return params['filename']
+    except KeyError:
```
ebd2: Why would this be missing if the repo has what we're looking for?

Do we create a more confusing error message later if we supply a filename here when the repo doesn't give us one and we guess wrong?

docs/presto-admin-commands.rst
zachyee: Yea, I like that better. I'll make the switch.

docs/presto-admin-commands.rst
zachyee: It doesn't right now. That will have to be added later when they can be automatically uploaded and are accessible. I think there's some issues right now because they have some additional features that shouldn't be publicly available. Matt said he would address this after this functionality is added.

prestoadmin/server.py
```diff
@@ -82,11 +86,191 @@ def new_sysnode_processor(node_info_rows):
...
+    print 'Downloading rpm from %s\n' \
+          'This can take a few minutes.' % url_response.geturl()
+
+    with open(download_file_path, 'wb') as local_file:
+        bytes_read = 0
+        block_size = 16 * 1024 * 1024
+        while True:
+            download_buffer = url_response.read(block_size)
+            if not download_buffer:
+                break
```
ebd2: What happens on a partial download?

prestoadmin/server.py
```diff
@@ -82,11 +86,191 @@ def new_sysnode_processor(node_info_rows):
...
+    """
+    try:
+        url_response = urllib2.urlopen(url)
+    except ValueError:
+        return None
+
+    if not _check_good_response_status(url_response.getcode()):
+        return None
+
+    download_file_name = _get_download_file_name(url_response, version)
```
ebd2: Ah. guessing doesn't matter.

prestoadmin/server.py
zachyee: HTTP makes no guarantee that the response will contain the `Content-Disposition` or the `Content-Length` headers. This catches the cases where the responses do not have those headers, so the program can continue.

prestoadmin/server.py
```diff
@@ -82,11 +86,191 @@ def new_sysnode_processor(node_info_rows):
...
+    """
+    Attempt to find or download the given rpm version.
+
+    Default to using Facebook rpm if 'f'aceboook and 't'eradata
+    are not specified in rpm_version.
+    """
+    if not _check_if_valid_version(rpm_version):
+        return None
+
+    if rpm_version[-1] != 'f' and rpm_version[-1] != 't':
```
ebd2: I don't think we guarantee that 't' will be the last component in the version. 0.148-t.1.1 is a version number that's been mentioned in #git-workflow.

prestoadmin/server.py
```diff
@@ -82,11 +86,202 @@ def new_sysnode_processor(node_info_rows):
...
+
+    The version is considered valid if it follows the format of
+    Semantic Versioning. That means it should have the form
+    X.Y.Z for Major.Minor.Patch. Z may not be applicable for
+    all rpm versions, so X.Y is also valid. X.Y.Z must all be numbers.
+
+    The version may be optionally followed by t (for teradata)
+    or f (for facebook), but no other letters are acceptable.
+    """
+    return re.match('^[0-9]+(\.[0-9]+){1,2}[tf]?$', rpm_version)
```
rschlussel: This isn't the right regex for 148t and later teradata versions.  148t looks like 0.148-t.x.y

prestoadmin/server.py
```diff
@@ -82,11 +86,202 @@ def new_sysnode_processor(node_info_rows):
...
+def _check_if_valid_version(rpm_version):
+    """
+    This function returns whether the input rpm_version is valid.
+
+    The version is considered valid if it follows the format of
+    Semantic Versioning. That means it should have the form
+    X.Y.Z for Major.Minor.Patch. Z may not be applicable for
+    all rpm versions, so X.Y is also valid. X.Y.Z must all be numbers.
+
+    The version may be optionally followed by t (for teradata)
```
rschlussel: I think appending f or t onto the version name to indicate where to download from is a bit hack-y given that it's never the actual last letter in a fb version, and is no longer the last letter in teradata versions.  I would rather have an optional argument to the command to indicate where to download from.

docs/presto-admin-commands.rst
rschlussel: +1

prestoadmin/server.py
```diff
@@ -82,11 +86,202 @@ def new_sysnode_processor(node_info_rows):
...
+
+
+def _find_or_download_most_recent_presto_rpm():
+    latest_rpm_url = 'https://repository.sonatype.org/service/local/artifact/maven' \
+                     '/content?r=central-proxy&g=com.facebook.presto' \
+                     '&a=presto-server-rpm&e=rpm&v=RELEASE'
+    return _find_or_download_rpm(latest_rpm_url)
+
+
+def _check_if_valid_version(rpm_version):
```
rschlussel: I don't think we need to check the version.  For teradata releases we'd need to have backwards compatibility for our different numbering schemes, and the whole thing could break if someone at some point decides a different versioning scheme would be better.  Also it still doesn't prevent you from specifying versions that don't exist.  I think the error message that would come up when you can't download the invalid version is just fine.

zachyee:
Addressed comments. You can now use `--rpm-source='facebook'` or `--rpm-source='teradata'`.

docs/presto-admin-commands.rst
```diff
@@ -392,9 +392,17 @@ server install
 **************
 ::
 
-    presto-admin server install <local_path> [--nodeps]
+    presto-admin server install <rpm_specifier> [--rpm-source] [--nodeps]
 
-This command copies the presto-server rpm from ``local_path`` to all the nodes in the cluster, installs it, deploys the general presto configuration along with tpch connector configuration. The ``local_path`` should be accessible by ``presto-admin``.
+This command takes in a parameter ``rpm_specifier``, which must be a string describing the presto-server rpm to be used.
+The string can be a local path to a previously downloaded rpm. The local path should be accessible by ``presto-admin``.
+The string can be a version number. In this case, this command will use the argument ``--rpm-source`` to determine where to download the rpm from. This command attempts to find a local rpm that matches the rpm that would be downloaded. If no match is found, it will attempt to download the rpm.
```
rschlussel: you should also note that the default value for --rpm-source is 'facebook' since it's an optional argument.

prestoadmin/main.py
```diff
@@ -348,6 +348,16 @@ def parser_for_options():
         help="default to serial execution method"
     )
 
+    advanced_options.add_option(
+        '--rpm-source',
+        action='store',
+        type='string',
```
ebd2: Add `choices=['facebook', 'teradata'] to reduce the error checking you need to do.

ebd2: I think that means the type needs to be `choice` too.

prestoadmin/main.py
```diff
@@ -565,6 +575,10 @@ def run_tasks(task_list):
                                  % name)
                 display_command(name, 2)
 
+            if state.env.rpm_source and name.strip() != 'server.install':
```
ebd2: I'd structure this the same as the above (nodeps) for consistency even though it's a list of one at the moment.

prestoadmin/server.py
```diff
@@ -82,11 +85,218 @@ def new_sysnode_processor(node_info_rows):
 CONNECTOR_INFO_SQL = 'select catalog_name from system.metadata.catalogs'
 _LOGGER = logging.getLogger(__name__)
 
+DOWNLOAD_DIRECTORY = main_dir
```
ebd2: This seems like a bad choice for anybody running with a small root filesystem. /tmp is probably a better choice. I'd argue that anybody downloading the RPM automatically probably isn't interested in keeping it around forever.

We should probably print a message saying where the RPM gets downloaded to.

rschlussel: Agree. It's also not that easy to find since when you install presto-admin, the main_dir becomes buried quite deep.

prestoadmin/server.py
```diff
@@ -82,11 +85,218 @@ def new_sysnode_processor(node_info_rows):
...
+    """
+    Args:
+        url:      The url of the presto rpm to be downloaded.
+        version:  An optional version number.
+                  If the server doesn't respond with the file name that is being
+                  requested, this allows the downloaded file to have the correct
+                  version attached to its name (presto-server-rpm-'version'.rpm)
+                  rather than the default name
+
+    If downloading the presto rpm at the given url would overwrite an existing rpm,
```
ebd2: At the very least, if this returns the path to an existing file, there needs to be a message to the user along the lines of "Using existing rpm ${full_path_to_rpm}. Delete the existing rpm to force a new download."

prestoadmin/server.py
```diff
@@ -82,11 +85,218 @@ def new_sysnode_processor(node_info_rows):
...
+
+    if not _check_good_response_status(url_response.getcode()):
+        return None
+
+    download_file_name = _get_download_file_name(url_response, version)
+    download_file_path = os.path.join(DOWNLOAD_DIRECTORY, download_file_name)
+
+    downloaded_rpm_path = _find_downloaded_rpm(download_file_name)
+    if downloaded_rpm_path and download_file_name != DEFAULT_RPM_NAME:
+        return downloaded_rpm_path
```
ebd2: What happens to the open http connection if we take this branch?

prestoadmin/server.py
```diff
@@ -82,11 +85,218 @@ def new_sysnode_processor(node_info_rows):
...
+
+    return _download_rpm(url_response, download_file_path)
+
+
+def _find_or_download_rpm_version_facebook(rpm_version):
+    download_url = 'http://search.maven.org/remotecontent?filepath=com/facebook/presto/' \
```
ebd2: Link the API docs, please: http://search.maven.org/#api

prestoadmin/server.py
```diff
@@ -82,11 +85,218 @@ def new_sysnode_processor(node_info_rows):
...
+    return _find_or_download_rpm(latest_rpm_url)
+
+
+def _find_or_download_latest_presto_rpm_teradata():
+    abort('Finding and downloading latest teradata presto rpm is currently not supported.\n'
+          'Try again after downloading the rpm and specifying the local path to the rpm')
+
+
+def _check_good_response_status(response_status):
+    return response_status == 200
```
ebd2: A 404 should result in an error along the lines of "Failed to find <facebook|terdata> version <version number>" but better worded :-)

prestoadmin/server.py
```diff
@@ -82,11 +85,218 @@ def new_sysnode_processor(node_info_rows):
...
+
+def _find_or_download_rpm_version_facebook(rpm_version):
+    download_url = 'http://search.maven.org/remotecontent?filepath=com/facebook/presto/' \
+                   'presto-server-rpm/' + rpm_version + '/presto-server-rpm-' + \
+                   rpm_version + '.rpm'
+    return _find_or_download_rpm(download_url, rpm_version)
+
+
+def _find_or_download_rpm_version_teradata(rpm_version):
+    abort('Finding and downloading teradata presto rpms by version is currently not supported.\n'
```
ebd2: Why do we expose this at all? Is there a plan in place to make this possible without going through the teradata website and filling in that form?

prestoadmin/server.py
```diff
@@ -104,15 +314,31 @@ def install(local_path):
...
+
+                        Before downloading an rpm, install will attempt to find a local
+                        copy with a matching version number to the requested rpm. If such
+                        a match is found, it will use the local copy instead of downloading
+                        the rpm again.
+
+        --rpm-source -  String specifying where the presto rpm should be downloaded from if
+                        necessary. By default, this is set to 'facebook'.
+
+        --nodeps -      (optional) Flag to indicate if server install
```
ebd2: Larger question than this PR: Should we expose this in the docstrings or leave it hidden? What's the use-case for this?

tests/product/standalone/presto_installer.py
```diff
@@ -48,15 +48,26 @@ def get_dependencies():
                 TopologyInstaller]
 
     def install(self, extra_configs=None, coordinator=None,
-                pa_raise_error=True):
+                pa_raise_error=True, rpm_specifier=None):
```
ebd2: Why not pass rpm_specifier into the constructor and handle the logic of deciding what to install there? That way the error checking for _detect_presto_rpm can stay where it was and the error gets raised at its source rather than sometime later.

prestoadmin/server.py
ebd2: Also: This is totally user-hostile for the case of a partial download or corrupted file.

tests/product/standalone/presto_installer.py
zachyee: I reverted these changes because product tests require you to download the presto rpm and testing the rpm download is done through unit tests, so the changes aren't needed.

prestoadmin/server.py
zachyee: I believe so. @mattsfuller

prestoadmin/server.py
zachyee: See `_find_downloaded_rpm()` starting at Line 155. I think it addressed these issues, although there might be a more elegant way to go about doing so.

prestoadmin/server.py
zachyee: If a 404 occurs and none of the other ways of finding the rpm work out, an `abort()` is called with a message similar to what you suggest. (Line 299)

prestoadmin/server.py
```diff
@@ -82,11 +84,226 @@ def new_sysnode_processor(node_info_rows):
...
+    If the rpm cannot be found, the function returns None.
+    If the rpm is found but corrupt, the function deletes the corrupt rpm and returns None.
+    If the rpm is found and is not corrupt, the function returns the path to the rpm.
+    """
+    download_file_path = os.path.join(DOWNLOAD_DIRECTORY, download_file_name)
+
+    sys.stderr = open(os.devnull, 'w')
+    try:
+        package.check_if_valid_rpm(download_file_path)
+        print 'Found exiting rpm at: %s\n' \
```
ebd2: typo: exiting -> existing

prestoadmin/server.py
ebd2: Works for me.

tests/product/standalone/presto_installer.py
```diff
@@ -109,7 +109,7 @@ def copy_presto_rpm_to_master(self, cluster=None):
     def _detect_presto_rpm():
         """
         Detects the Presto RPM in the main directory of presto-admin.
-        Returns the name of the RPM, if it exists, else returns None.
+        Returns the name of the RPM, if it exists, else raises an OSError.
```
ebd2: Nice catch

prestoadmin/server.py
ebd2: How about adding a warn() with the specific URL we tried to aid troubleshooting? Or log it only. Either seems OK.

tests/unit/test_server.py
```diff
@@ -40,28 +40,121 @@ class TestInstall(BaseUnitCase):
...
+        rpm_specifier = "/any/path/rpm"
+        mock_find_local_rpm.return_value = rpm_specifier
+        env.rpm_source = 'facebook'
+        server.install(rpm_specifier)
+        mock_check_rpm.assert_called_with(rpm_specifier)
         mock_execute.assert_called_with(server.deploy_install_configure,
-                                        local_path, hosts=get_host_list())
+                                        rpm_specifier, hosts=get_host_list())
+
+    @patch('prestoadmin.server._find_local_presto_rpm')
```
ebd2: Here and throughout, this is kind of a weird use of mock objects. Generally, you want to try to mock calls to things that don't exist in testing (some prod-only resource), are expensive to call (downloading the world), or are hard to force into a specific error state (return a 451 error from a web server).

The way this is mocked is pretty tightly coupled with the current implementation, which adds a lot of maintenance cost. A more general approach might be along the following lines:

0. Mock os.path.isfile so that find_local_presto_rpm returns the value you pass into server.install
1. Mock package.local so that check_if_valid_rpm thinks the rpm is good.
2. Mock urllib2.urlopen and assert we don't call it.
3. Leave the execute mock as it is.

zachyee: My rationale for creating all of the mock objects was that the rpm_specifier can take on a few different forms and I wanted to use the mocks to control which form was used for the test. For instance, if I made rpm_specifier = 0.148, I want it to be interpreted as a version, so I make sure that finding it locally fails. If I want it to check the download URL, I make sure that it doesn't find a local downloaded copy. In both these cases, I have to mock those functions so I can set their return values to signal that they failed. This way, I make sure I'm testing whatever single function I'm trying to test. Isn't that what I want from a unit test?

What you're suggesting is mocking the lower-level functions in the higher-level ones. If we wanted to do that, wouldn't it be better to have these tests mock the higher level functions like `find_local_presto_rpm` and `check_if_valid_rpm`, and then create separate unit tests for these higher level functions where we mock `os.path.isfile` and `package.local`?

docs/presto-admin-commands.rst
```diff
@@ -392,9 +392,18 @@ server install
 **************
 ::
 
-    presto-admin server install <local_path> [--nodeps]
-
-This command copies the presto-server rpm from ``local_path`` to all the nodes in the cluster, installs it, deploys the general presto configuration along with tpch connector configuration. The ``local_path`` should be accessible by ``presto-admin``.
+    presto-admin server install <rpm_specifier> [--rpm-source] [--nodeps]
+
+This command takes in a parameter ``rpm_specifier``, which must be a string describing the presto-server rpm to be used.
+The string can be 'latest'. This finds a local copy or downloads of the latest version of the presto rpm.
```
ebd2: Suggest: 'The parameter can be one of the following forms, listed in order of decreasing precedence:'
and omit 'The string can be' from each form.

docs/presto-admin-commands.rst
```diff
@@ -392,9 +392,18 @@ server install
...
+The string can be a url. This finds a local copy or downloads the presto rpm found at the given url.
```
ebd2: local copy via a 'file:///' URL? If so, I'd suggest omitting that; it'll only confuse people who don't know about file:// URLs. People who know about them are probably savvy enough to figure them out themselves.

If not referring to file:// URLs, please explain further.

docs/presto-admin-commands.rst
```diff
@@ -392,9 +392,18 @@ server install
...
+The string can be a version number. This finds a local copy or downloads the presto rpm of the specified version.
```
ebd2: Suggest: give an example version number.

ebd2: Nevermind, there are examples below.

prestoadmin/server.py
```diff
@@ -80,11 +84,254 @@ def new_sysnode_processor(node_info_rows):
...
+
+    @staticmethod
+    def _check_rpm_uncorrupted(rpm_path):
+        old_stderr = sys.stderr
+        sys.stderr = open(os.devnull, 'w')
+        try:
+            package.check_if_valid_rpm(rpm_path)
+        except SystemExit as system_exit:
+            if 'Corrupted RPM' in system_exit.message:
+                print('Removing corrupted rpm at: %s' % rpm_path)
```
ebd2: This should probably use fabric's warn() function. It should probably also say something like "Trying to remove ..."

prestoadmin/server.py
```diff
@@ -80,11 +84,254 @@ def new_sysnode_processor(node_info_rows):
...
+        self.local_path = local_path
+
+    @staticmethod
+    def _check_rpm_uncorrupted(rpm_path):
+        old_stderr = sys.stderr
+        sys.stderr = open(os.devnull, 'w')
+        try:
+            package.check_if_valid_rpm(rpm_path)
+        except SystemExit as system_exit:
+            if 'Corrupted RPM' in system_exit.message:
```
ebd2: This could be problematic if rpm is internationalized/localized. We should try to do this on exit code alone if at all possible.

prestoadmin/server.py
```diff
@@ -80,11 +84,254 @@ def new_sysnode_processor(node_info_rows):
...
+                print('Removing corrupted rpm at: %s' % rpm_path)
+                try:
+                    os.remove(rpm_path)
+                except OSError:
+                    pass
```
ebd2: warn('Life is terrible, delete it yourself')

prestoadmin/server.py
```diff
@@ -80,11 +84,254 @@ def new_sysnode_processor(node_info_rows):
...
+
+
+class LocalPrestoRpmFinder:
+    def __init__(self, local_path):
+        self.local_path = local_path
+
+    @staticmethod
+    def _check_rpm_uncorrupted(rpm_path):
+        old_stderr = sys.stderr
+        sys.stderr = open(os.devnull, 'w')
```
ebd2: Comment why we're doing this please.

tests/product/test_server_install.py
```diff
@@ -193,7 +202,7 @@ def test_install_with_java8_home(self):
         expected = installed_all_hosts_output
 
         actual = cmd_output.splitlines()
-        self.assertEqual(sorted(expected), sorted(actual))
+        self.assertRegexpMatchesLineByLine(actual, expected)
```
ebd2: Why these changes? I've long thought that this was to deal with the fact that there are no guarantees about what order of the actual output will be since there are 3-4 processes running in parallel.

tests/unit/test_server.py
```diff
@@ -41,28 +42,290 @@ class TestInstall(BaseUnitCase):
...
+        self.check_find_valid_rpm_returns_absolute_path(is_absolute_path=True)
+
+    def test_check_find_valid_rpm_at_relative_path_returns_absolute_path(self):
+        self.check_find_valid_rpm_returns_absolute_path(is_absolute_path=False)
+
+    @patch('prestoadmin.server.urllib2.urlopen')
+    def check_content_length(self, mock_urlopen, is_header_present):
+        url_response = MagicMock()
+        if is_header_present:
+            url_response.info.return_value = {'Content-Length': '123'}
```
ebd2: Slick!

tests/product/test_server_install.py
zachyee: Yea that's right. Internally, assertRegexpMatchesLineByLine sorts expect and actual and does a regexp comparison line by line (rather than just string equality as before). I made this change because I added some regex strings to the expected output, so I couldn't just test on string equality.

ebd2: Gotcha, thanks.

ebd2:
LGTM

PR #208, opened 2016-07-21
Add comments about supported images

zachyee:


zachyee:
@ebd2

ebd2:
Great explanation why it is as it is. LGTM

PR #210, opened 2016-07-22
Make ensure_containers_started opt-out.

ebd2:
Previous opt-in behavior caused problems when we change the naming of our
images. We stopped waiting for containers to fully start, and had a mess of
intermittent problems because of it. The new behavior ensures that if we change
the naming of an opted-out image, we'll see failures immediately and can
correct them.

### DIFF
--- a/tests/docker_cluster.py
+++ b/tests/docker_cluster.py
@@ -18,25 +18,38 @@ Test writers should use this module for all of their docker related needs
 and not directly call into the docker-py API.
 """
 
+import errno
+import os
+import shutil
 import subprocess
 import sys
-import shutil
-import os
-import errno
 import uuid
 from time import sleep
 
 from docker import Client
 from docker.errors import APIError
 from docker.utils.utils import kwargs_from_env
+from retrying import retry
 
 from prestoadmin import main_dir
 from tests.base_cluster import BaseCluster
 from tests.product.constants import \
-    DEFAULT_DOCKER_MOUNT_POINT, DEFAULT_LOCAL_MOUNT_POINT, \
-    BASE_TD_IMAGE_NAME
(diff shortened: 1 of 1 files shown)


tests/docker_cluster.py
```diff
@@ -249,31 +261,23 @@ def _add_hostnames_to_slaves(self):
                 % additions_to_etc_hosts
             )
 
+    @retry(stop_max_delay=_DOCKER_START_TIMEOUT, wait_fixed=_DOCKER_START_WAIT)
     def _ensure_docker_containers_started(self, image):
-        centos_based_images = [BASE_TD_IMAGE_NAME]
-
-        timeout = 0
-        is_host_started = {}
+        host_started = {}
```
ebd2: Sorry about the churn. TIL dict comprehensions don't work in 2.6 :-(

tests/docker_cluster.py
```diff
@@ -18,26 +18,39 @@
...
-    DEFAULT_DOCKER_MOUNT_POINT, DEFAULT_LOCAL_MOUNT_POINT, \
-    BASE_TD_IMAGE_NAME
+    DEFAULT_DOCKER_MOUNT_POINT, DEFAULT_LOCAL_MOUNT_POINT
 
 DIST_DIR = os.path.join(main_dir, 'tmp/installer')
 
+_DOCKER_START_TIMEOUT = 30000
+_DOCKER_START_WAIT = 1000
+
+ENSURE_STARTED_OPT_OUT_IMAGES = [
```
rschlussel: Could you rename this?  It took me a while to figure out what you meant by opt out. What about NON_CENTOS_IMAGES?  The point is just that we don't check about centos services because it's not a centos container.

rschlussel:
looks good % rename

ebd2:
Addressed feedback and merged.

PR #211, opened 2016-07-22
Use retrying instead of our own implementation.

ebd2:
Our implmentation does not account for time spent trying stuff. When we're
trying slow things that fail, this ends up being the lion's share of the time.


rschlussel:
looks good

ebd2:
Thanks. I'll push Monday after making sure the tses pass.

PR #216, opened 2016-07-28
Fix timeout in status product tests

ebd2:
After changing to retrying.retry, we're hitting errors because retrying
accounts for time spent actually trying stuff. In some cases, getting the
status takes over 2 minutes, which is a problem in and of itself, but we need
to stop the emails from Jenkins. Bump the timeout to 3 minutes, which allows
time for a second attempt.


cawallin:
lgtm

PR #217, opened 2016-08-02
Allow non-root user in config.json to work

cawallin:
When we updated presto-admin to make the config files owned by
presto:presto, because they have passwords, we neglected
to use sudo to access the files (which is necessary if the
user in config.json is not either root or the presto user)

### DIFF
--- a/prestoadmin/collect.py
+++ b/prestoadmin/collect.py
@@ -44,6 +44,7 @@ from util.constants import PRESTOADMIN_LOG_DIR
 
 
 TMP_PRESTO_DEBUG = '/tmp/presto-debug/'
+TMP_PRESTO_DEBUG_REMOTE = '/tmp/presto-debug-remote'
 OUTPUT_FILENAME_FOR_LOGS = '/tmp/presto-debug-logs.tar.bz2'
 OUTPUT_FILENAME_FOR_SYS_INFO = '/tmp/presto-debug-sysinfo.tar.bz2'
 PRESTOADMIN_LOG_NAME = 'presto-admin.log'
@@ -108,7 +109,7 @@ def get_files(remote_path, local_path):
     _LOGGER.debug('local path used ' + path_with_host_name)
 
     try:
-        get(remote_path, path_with_host_name, True)
+        get(remote_path, path_with_host_name, use_sudo=True)
     except SystemExit:
         warn('remote path ' + remote_path + ' not found on ' + env.host)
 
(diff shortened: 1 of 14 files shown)


cawallin:
This still needs automated tests -- but it was tested manually. It also must be tested on EMR before it is merged.

prestoadmin/collect.py
```diff
@@ -209,10 +210,10 @@ def system_info():
 
 def get_system_info(download_location):
 
-    if not exists(TMP_PRESTO_DEBUG):
-        run("mkdir " + TMP_PRESTO_DEBUG)
+    if not exists(TMP_PRESTO_DEBUG_REMOTE):
```
akshatnair: this seems unrelated

cawallin: it actually is related -- on the master, /tmp/presto-debug has already been created and is owned by root. when presto-admin logs into all the nodes and tries to create /tmp/presto-debug/sysinfo.txt with a non-root user, it fails

cawallin:
Tests added

prestoadmin/collect.py
akshatnair: but why was it already created as root? May be leftover from some other code?  Shouldn't all these be created with the same user as this is part of collect logs collect system_info

cawallin: it's created by the user that's running presto-admin, which right now always is root since you have to run presto-admin as sudo.

collect system_info is the only of the "collect" commands that needs to write out a file when executing a command on the remote nodes, and that remote command is run as the username you specify for sudo (which may not be root)

tests/unit/test_server.py
```diff
@@ -323,14 +323,14 @@ def test_collect_node_information(self, mock_ext_ip, mock_service,
         self.assertEqual(('IP4', False, ''),
                          server.collect_node_information())
 
-    @patch('prestoadmin.server.run')
+    @patch('prestoadmin.server.sudo')
     def test_get_external_ip(self, mock_nodeuuid):
```
akshatnair: why is this called mock_uuid? and not mock_run and now mock_sudo?

akshatnair:
lgtm mod small comments.

tests/unit/test_server.py
cawallin: it's the command to get the node uuid, seems fine to keep it as is

ebd2:
Did you look at the use of `put` in script.py and the several places that use `put_secure`?

Apart from that question, this looks good to me.

cawallin:
I was assuming that you'd put the script in a place that the user has access to, it's run via sudo. I covered all of the places that use put_secure.

Works fine on EMR, I'm going to merge this.

PR #218, opened 2016-08-02
Add release command to setup.py

zachyee:
In order to automate releasing to github and pypi

Currently uses basic authentication (username/password) to release to github. An alternative would be to use OAuth tokens: https://developer.github.com/v3/oauth/

It also currently can only release to pypitest, but adding pypi to `.pypirc` will allow it to release to pypi.
It uploads a source distribution (default for unix is .tar.gz).


release.py
```diff
@@ -0,0 +1,440 @@
...
+    user_options = [('github', None,
+                     'boolean flag indicating if a release should be created for github'),
+                    ('pypi', None,
+                     'boolean flag indicating if a release should be created for pypi'),
+                    ('all', None,
+                     'boolean flag indicating if a release should be created for github and pypi')]
+
+    @staticmethod
+    def get_latest_release():
+        latest_release = subprocess.check_output(['curl', '--silent', REPOSITORY_API + '/releases/latest'])
```
ebd2: Why not urllib2 or something like that?

release.py
```diff
@@ -0,0 +1,440 @@
...
+    def check_branch_remote_exists(local_branch):
+        for remote_branch in release.get_remote_branches():
+            if local_branch == remote_branch['name']:
+                print 'Local branch %s exists remotely' % local_branch
+                return
+        exit('Local branch %s does not exist remotely.' % local_branch)
+
+    @staticmethod
+    def check_branch_up_to_date(local_branch):
+        status = subprocess.check_output(['git', 'status'])
```
ebd2: Same question but for the git module

ebd2: See https://github.com/kokosing/docker-release/blob/master/docker_release/main.py for an example

setup.py
```diff
@@ -109,14 +110,15 @@ def get_custom_rc_file(self):
     classifiers=[
         'Development Status :: 2 - Pre-Alpha',
         'Intended Audience :: Developers',
-        'License :: OSI Approved :: APLv2 License',
+        'License :: OSI Approved :: Apache Software License',
```
ebd2: Dumb question from the ignorant: What is APLv2 and Apache Software License? Neither of them appear listed here: https://www.apache.org/licenses/

@mattsfuller: Can you confirm what this should be? You spend more time talking to the lawyers than we do.

ebd2:
Looks basically OK to me, modulo the question about about going through subprocess rather than python modules and the license question.

release.py
```diff
@@ -0,0 +1,440 @@
...
+import os
+import re
+import subprocess
+
+try:
+    from setuptools import Command
+except ImportError:
+    from distutils.core import Command
+
+REPOSITORY_API = 'https://api.github.com/repos/zachyee/presto-admin'
```
ebd2: Testing URL? Change before pushing to master?

setup.py
zachyee: Would be good to confirm. Also here's a list of valid classifiers: https://pypi.python.org/pypi?%3Aaction=list_classifiers

release.py
```diff
@@ -0,0 +1,490 @@
...
+    from distutils.core import Command
+
+REPOSITORY_API = 'https://api.github.com/repos/prestodb/presto-admin'
+CURRENT_DIRECTORY = os.path.dirname(os.path.realpath(__file__))
+RELEASE_LIST_DOC_PATH = os.path.join(CURRENT_DIRECTORY, 'docs/release.rst')
+RELEASE_DOCS_DIRECTORY = os.path.join(CURRENT_DIRECTORY, 'docs/release/')
+DIST_DIRECTORY = os.path.join(CURRENT_DIRECTORY, 'dist')
+
+
+class release(Command):
```
ebd2: Most of the methods in here are static. Maybe it makes sense to break out github and pypi files that handle the inner workings of pushing to github and pypi and invoke that functionality here? That might also have the benefit of helping to make clear what the public interfaces are and what functions are internal implementation details.

setup.py
mattsfuller: It should be Apache License Version 2

zachyee:
Refactored into separate classes

release.py
```diff
@@ -0,0 +1,488 @@
...
+
+    @staticmethod
+    def _get_version_field_value(version_fields, index):
+        try:
+            return version_fields[index]
+        except IndexError:
+            # The field value was omitted for the version
+            return 0
+
+    def _set_semantic_version_fields(self):
```
ebd2: Nit: this should probably be inline in the constructor. It isn't used elsewhere, and it's weird that it operates by side-effect like this.

release.py
```diff
@@ -0,0 +1,488 @@
...
+try:
+    from setuptools import Command
+except ImportError:
+    from distutils.core import Command
+
+GITHUB_REPOSITORY_API = 'https://api.github.com/repos/prestodb/presto-admin'
+CURRENT_DIRECTORY = os.path.dirname(os.path.realpath(__file__))
+
+
+class SemanticVersionHelper(object):
```
ebd2: The object represents a specific version. Proposed name: SemanticVersion

release.py
```diff
@@ -0,0 +1,488 @@
...
+            # The field value was omitted for the version
+            return 0
+
+    def _set_semantic_version_fields(self):
+        version_fields = self.version.split('.')
+        self.major_version = self._get_version_field_value(version_fields, 0)
+        self.minor_version = self._get_version_field_value(version_fields, 1)
+        self.patch_version = self._get_version_field_value(version_fields, 2)
+
+    def __lt__(self, other):
```
ebd2: Please imlement __eq__ and use @total_ordering on the class.

release.py
```diff
@@ -0,0 +1,488 @@
...
+        or 0s. For instance, bumping 0.1.2's major version can result in 1, 1.0, or 1.0.0.
+        """
+        major_bumps = self._get_acceptable_major_version_bumps()
+        minor_bumps = self._get_acceptable_minor_version_bumps()
+        patch_bumps = self._get_acceptable_patch_version_bumps()
+        return major_bumps + minor_bumps + patch_bumps
+
+
+class ReleaseFetcher:
+    def __init__(self, directory, github_api_directory):
```
ebd2: suggest: github_api_path; it isn't really a directory in the sense that it's a container for files.

release.py
```diff
@@ -0,0 +1,488 @@
...
+                    exit('Expected %s to begin with "%s"' % (file_path, string_contained))
+
+            return file_contents
+
+    def _confirm_version_changed_in_file(self, file_path, tag_name):
+        contents = self._get_and_check_release_file(file_path)
+        for line in contents.splitlines():
+            if 'version' in line and '=' in line:
+                if tag_name not in line:
+                    exit('Version has not been updated to %s in %s' % (tag_name, file_path))
```
ebd2: Include a line number to aid problem solving. The check for a version is a little coarse, which is OK for internal tools, but we should make it easy to know what the problem is if this code finds one.

ebd2: Actually, it would probably be preferable to extract a version constant in setup.py, and then `from prestoadmin import version` and `from setup import version`, which lets us rely on the python parser.

release.py
```diff
@@ -0,0 +1,488 @@
...
+        self.password = None
+        self.tag_name = None
+        self.release_name = None
+        self.target_commitish = None
+        self.name = None
+        self.body = None
+        self.is_draft = 'false'
+        self.is_prerelease = 'false'
+
+    def _user_set_username(self):
```
ebd2: Here and below, suggest name `prompt_username`

release.py
```diff
@@ -0,0 +1,488 @@
...
+        """
+        self.target_commitish = self.release_fetcher.get_and_check_target_commitish()
+        self.tag_name = self.release_fetcher.get_and_check_tag()
+        self.release_validator.confirm_all_release_file_changes(self.tag_name)
+        self.body = self.release_fetcher.get_body_from_release_notes(self.tag_name)
+        self.body = self._make_multiline_string_json_friendly(self.body)
+        self.release_name = 'Release ' + self.tag_name
+
+    @staticmethod
+    def _make_multiline_string_json_friendly(multiline_string):
```
ebd2: suggest `escape_newlines` for the name.

release.py
```diff
@@ -0,0 +1,488 @@
...
+                  self.body, self.is_draft, self.is_prerelease)
+
+    @staticmethod
+    def _successful_post_request(status):
+        if status == 201:
+            return True
+        else:
+            return False
+
+    def _send_post_request(self, url, data, authorization_string, content_type, content_length):
```
ebd2: The GET/POST stuff probably belongs in a new module under `prestoadmin/util`. It's more general than just talking to github.

zachyee:
Addressed comments.

Created a util directory with `util/semantic_version.py` and `util/http.py`.

Created `prestoadmin/_version.py` which holds the current version. Unfortunately, setup.py should not import directly from `prestoadmin` because it may (and does) contain third party dependencies. Instead it reads the file, so it doesn't have to be parsed. This is a workaround other modules use, even though it's not pretty. Comments were added to note this.

ebd2:
Looks good to me except for the license.

PR #219, opened 2016-08-09
Remove presto group when checking config files

cawallin:
Some sudoers files don't allow root to run commands as other groups.
The presto group isn't necessary when grepping for config values,
just the user, so we remove it.

Tested on the Teradata Hadoop Appliance and EMR, as well as docker.


ebd2:
LGTM

PR #220, opened 2016-08-09
Release 1.4.1

cawallin:


zachyee:
Add `release/release-1.4.1` to `docs/release.rst`

cawallin:
Good call, thanks

cawallin:
updated

rschlussel:
should this be release 1.5 because of the added support for 0.149+?  Seems like a significant change (definitely bigger than the difference between 1.2 and 1.3)

PR #235, opened 2016-09-06
Guard against no RepoTags and small refactor

petroav:
- Some of the docker images I have locally do not have
  any RepoTags, causing a failure in the code checking if certain
  images exist. Change the checking code to guard against this
  scenario.
- Mark some methods as static.


ebd2:
LGTM

PR #236, opened 2016-09-07
Edit ConfigurableCluster & miscellaneous cleanups

petroav:
- Cache the Presto RPM instead of uploading it every time.
This isn't a problem with Docker because an "upload" is just
a file copy. However, with real clusters there's significant
latency involved in upload the RPM.
- Add key based authentication to the cluster
- Make the default mount point configurable
- Parametrize the install-admin.sh script
- Remove unused tests/product/__init__.py file used to check
for the presence of the ConfigurableCluster YAML file
- Move the run_installers() method to base_product_case.py


petroav:
Smoke test passed locally

tests/configurable_cluster.py
```diff
@@ -48,7 +48,11 @@ class ConfigurableCluster(BaseCluster):
     slaves: ['172.16.1.11', '172.16.1.12', '172.16.1.13']
     user: root
     teardown_existing_cluster: true
+    key_path: /path/to/cluster-key.pem
+    mount_point: /home/ec2-user/presto-admin
```
ebd2: This seems awfully specific.

tests/configurable_cluster.py
```diff
@@ -99,6 +107,7 @@ def tear_down(self):
             service presto stop
```
ebd2: How about the rpm_cache_dir and mount_point?

tests/configurable_cluster.py
```diff
@@ -140,6 +149,11 @@ def get_down_hostname(self, host_name):
 
     def exec_cmd_on_host(self, host, cmd, user=None, raise_error=True,
                          tty=False):
+        # Ensure command is invoked with sudo since EMR's login user is ec2-user
+        # If sudo is already present in the command then no error will occur
+        # as arbitrary nesting of sudo is allowed
```
ebd2: This give me shivers with regards to quoting. Have you tried this out with some fancy quoting and made sure that the command line doesn't get interpreted multiple times. It seems like it shouldn't, but confirmation would be good.

ebd2: Also, base_product_case.run_prestoadmin uses this. This will mean that we're always running presto-admin as sudo, which will mask any errors we want to ferret out in our transition to not requring root.

tests/configurable_cluster.py
```diff
@@ -148,11 +162,11 @@ def exec_cmd_on_host(self, host, cmd, user=None, raise_error=True,
             host = self.all_hosts()[index]
         ssh = paramiko.SSHClient()
         ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-        ssh.connect(host, username=user, password=self.password,
+        ssh.connect(host, username=user, key_filename=self.key_path,
```
ebd2: Hooray!

tests/configurable_cluster.py
petroav: Can you give me an example and I'll try it out?

tests/base_cluster.py
```diff
@@ -117,7 +117,7 @@ def exec_cmd_on_host(self, host, cmd, user=None, raise_error=True,
         pass
 
     @abc.abstractmethod
-    def run_script_on_host(self, script_contents, host):
+    def run_script_on_host(self, script_contents, host, tty=True):
```
ebd2: Can you please braindump your tty knowledge here so we capture it for posterity?

tests/configurable_cluster.py
```diff
@@ -194,29 +208,46 @@ def copy_to_host(self, source_path, host, dest_path=None):
             dest_path = os.path.join(self.mount_dir,
                                      os.path.basename(source_path))
         self.exec_cmd_on_host(host, 'mkdir -p ' + os.path.dirname(dest_path))
+        self.exec_cmd_on_host(host, 'chown {0}:{0} {1}'.format(self.user, os.path.dirname(dest_path)))
```
ebd2: Prefer dict-based formatting to make it less likely that somebody adds a parameter and screws up the positions.

tests/configurable_cluster.py
```diff
@@ -194,29 +208,46 @@ def copy_to_host(self, source_path, host, dest_path=None):
...
 
         ssh = paramiko.SSHClient()
         ssh.set_missing_host_key_policy(paramiko.AutoAddPolicy())
-        ssh.connect(host, username=self.user, password=self.password,
+        ssh.connect(host, username=self.user, key_filename=self.key_path,
                     timeout=180)
+
+        # Upload to dummy location because paramiko doesn't allow SFTP using
```
ebd2: Maybe note that fabric uses the same solution?

tests/configurable_cluster.py
```diff
@@ -194,29 +208,46 @@ def copy_to_host(self, source_path, host, dest_path=None):
...
+
+        # Move to final location using sudo
+        self.exec_cmd_on_host(host, 'mv {0} {1}'.format(dummy_path, dest_path))
+
+        # Remove dummy path directory
+        self.exec_cmd_on_host(host, 'rm -rf {0}'.format(os.path.dirname(dummy_path)))
+
         ssh.close()
 
+    # Since ConfigurableCluster is configured using external IPs, those act as
```
ebd2: Does it work with hostnames too? If so, what are the requirements for those being resolveable? From the machine running the tests only, or from within the cluster as well?

tests/product/__init__.py
```diff
@@ -1,24 +0,0 @@
...
-    if config_name:
-        return config_name[0]
-    else:
-        return None
-
-
-def setup_package():
-    if check_for_cluster_config():
-        global user_password
-        user_password = getpass.getpass('Password for cluster: ')
```
ebd2: Hooray!

tests/product/base_product_case.py
```diff
@@ -148,6 +149,20 @@ def setup_cluster(self, bare_image_provider, cluster_type):
             else:
                 raise RuntimeError("Docker images have not been created")
 
+    @staticmethod
+    def run_installers(cluster, installers, testcase):
```
ebd2: _run_installers? We shouldn't be using this outside of here, right?

ebd2: Oh, we use it in image_builder. Please comment that this shouldn't be used directly in test cases.

tests/product/prestoadmin_installer.py
```diff
@@ -51,14 +52,11 @@ def install(self, cluster=None, dist_dir=None):
...
-        cluster.exec_cmd_on_host(
-            cluster.master,
-            'chmod +x ' + cluster.mount_dir + "/install-admin.sh"
-        )
-        cluster.exec_cmd_on_host(
-            cluster.master, cluster.mount_dir + "/install-admin.sh")
+        with open(LOCAL_RESOURCES_DIR + "/install-admin.sh", 'r') as file_obj:
+            script = ''.join(file_obj.readlines())
+
+        script = script.format(cluster.mount_dir)
```
ebd2: Prefer dict-based formatting here where the format string and args are in 2 totally different places.

tests/product/prestoadmin_installer.py
```diff
@@ -51,14 +52,11 @@ def install(self, cluster=None, dist_dir=None):
...
-        cluster.copy_to_host(
-            LOCAL_RESOURCES_DIR + "/install-admin.sh", cluster.master)
-        cluster.exec_cmd_on_host(
-            cluster.master,
-            'chmod +x ' + cluster.mount_dir + "/install-admin.sh"
-        )
-        cluster.exec_cmd_on_host(
-            cluster.master, cluster.mount_dir + "/install-admin.sh")
+        with open(LOCAL_RESOURCES_DIR + "/install-admin.sh", 'r') as file_obj:
+            script = ''.join(file_obj.readlines())
```
ebd2: script = file_obj.read()?

tests/configurable_cluster.py
ebd2: touch "foo "'"'" bar"
bash -c "foo "'"'" bar"

I think you need to test both.

petroav: The first command you suggested works but the second doesn't. However, the second doesn't even if the command is executed without sudo so I'm not introducing any regressions and I won't debug further.

As for the point of invoking sudo for everything what I'll do is make sudo configurable so that when we start the "no sudo" work it can be turned off.

tests/configurable_cluster.py
petroav: This set of commands gets run before every test so deleting the rpm_cache_dir defeats the purpose of that directory. I also can't think of a global setUp()/teardown() method that could delete this directory. As for mount_dir, yes that gets deleted, see the last line of this command.

tests/base_cluster.py
petroav: I'll add a comment on the method.

tests/configurable_cluster.py
petroav: Done

tests/configurable_cluster.py
petroav: Done

tests/configurable_cluster.py
petroav: The cluster.master field is supposed to return an IP or hostname that is externally accessible from anywhere. Of course because we directly fetch that variable, there's no way to enforce it between implementations (that should change with a refactor I have in mind).

cluster.master should be resolvable from the machine running the tests and in the case of configurable cluster that happens to be my machine/jenkins which will use the external interface.

tests/product/base_product_case.py
petroav: Yes, good call.

tests/product/prestoadmin_installer.py
petroav: Yes, good call.

tests/product/prestoadmin_installer.py
petroav: I think that's less convenient. The whole code block would look like this:

```
file_obj = open(path)
script = file_obj.read(hardcoded_size)
file_obj.close()
```

The `with` context manager closes the file for me and I don't have to hardcode a size with the current version.

petroav:
Thanks for the review, I pushed a fixup commit that addresses your comments.

tests/product/prestoadmin_installer.py
petroav: Actually, you're right, with `read()`, I don't have to join the lines after. Changed.

tests/configurable_cluster.py
ebd2: Sounds reasonable.

tests/configurable_cluster.py
```diff
@@ -193,30 +207,51 @@ def copy_to_host(self, source_path, host, dest_path=None):
...
         sftp = ssh.open_sftp()
-        sftp.put(source_path, dest_path)
+        sftp.put(source_path, dummy_path)
         sftp.close()
+
+        # Move to final location using sudo
+        self.exec_cmd_on_host(host, 'mv {source} {dest}'.format(source=dummy_path, dest=dest_path))
+
+        # Remove dummy path directory
+        self.exec_cmd_on_host(host, 'rm -rf {dir}'.format(dir=os.path.dirname(dummy_path)))
```
ebd2: This seems like a good candidate for a context manager, but I'm fine with it like this for test code.

tests/product/base_product_case.py
```diff
@@ -205,7 +222,7 @@ def run_prestoadmin(self, command, raise_error=True, cluster=None,
         command = self.replace_keywords(command, cluster=cluster, **kwargs)
         return cluster.exec_cmd_on_host(
             cluster.master,
-            "/opt/prestoadmin/presto-admin %s" % command,
+            "/opt/prestoadmin/presto-admin --user {0} {1}".format(cluster.user, command),
```
ebd2: keyword formatting.

tests/configurable_cluster.py
```diff
@@ -64,10 +68,14 @@ def __init__(self, config_filename):
         self.internal_master = 'master'
         self.internal_slaves = ['slave1', 'slave2', 'slave3']
         self.user = config['user']
-        from tests.product import user_password
-        self.password = user_password
+
+        self.key_path = config['key_path']
+        if not os.path.exists(self.key_path):
+            raise Exception('Key path specified {0} does not exist.'.format(self.key_path))
```
ebd2: keyword formatting

ebd2:
LGTM mod the nits about keyword formatting.

tests/product/base_product_case.py
petroav: Done.

PR #239, opened 2016-09-08
New dummy-rpm that was built with a verion independent RPM

petroav:


petroav:
I guess you can't really review much since it's a binary file. I ran all the upgrade related tests on EMR and they pass, will do so with Docker as well before pushing.

rschlussel:
lgtm.  fix the typo in the commit message.

PR #254, opened 2016-09-14
Distinguish clean for test and test-all

rschlussel:
'make clean-test' should only get rid of the artifacts from 'make test'.
'make clean-test-all' will get rid of the containers and other artifacts
from 'make test-all'.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 .PHONY: clean-all clean clean-eggs clean-build clean-pyc clean-test-containers clean-test \
 	clean-docs lint smoke test test-all test-images test-rpm docker-images coverage docs \
-	open-docs release release-builds dist dist-online dist-offline wheel install precommit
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -60,10 +61,13 @@ clean-test:
 	rm -fr .tox/
 	rm -f .coverage
 	rm -fr htmlcov/
+
+clean-test-all: clean-test
 	rm -fr tmp
 	for image in $$(docker images | awk '/teradatalabs\/pa_test/ {print $$1}'); do docker rmi -f $$image ; done
 	@echo "\n\tYou can kill running containers that caused errors removing images by running \`make clean-test-containers'\n"
 
+
```
ebd2: spurious change?

rschlussel: yup.  thanks

ebd2:
LGTM

PR #256, opened 2016-09-14
Make travis only run product tests on python 2.7

rschlussel:
Since we want to exclude most of the tests from being run on 2.6, it seemed clearer to explicitly write out what we did want to run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 language: python
-python:
-  - "2.6"
-  - "2.7"
+python: "2.7"
 sudo: required
+dist: trusty
 services:
   - docker
 env:
@@ -21,6 +20,7 @@ env:
 install:
   - pip install --upgrade pip==6.1.1
   - pip install -r requirements.txt
+  - pip install tox-travis
 before_script:
   - make docker-images
   - make presto-server-rpm.rpm
@@ -47,8 +47,10 @@ script:
       nosetests --with-timer --timer-ok 60s --timer-warning 300s -a '!quarantine,!offline_installer' ${LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO};
     elif [ -v OTHER_TESTS ]; then
       make clean lint dist docs test-rpm &&
-      nosetests -s tests.unit &&
-      nosetests -s tests.integration;
+      tox -e py26 -- -s tests.unit &&
+      tox -e py26 -- -s tests.integration;
+      tox -e py27 -- -s tests.unit &&
+      tox -e py27 -- -s tests.integration;
     else
       echo "Unknown test" &&
       exit 1;
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,7 +46,7 @@ script:
       make test-images &&
       nosetests --with-timer --timer-ok 60s --timer-warning 300s -a '!quarantine,!offline_installer' ${LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO};
     elif [ -v OTHER_TESTS ]; then
-      make clean lint dist docs test-rpm &&
+      make clean lint dist docs &&
       tox -e py26 -- -s tests.unit &&
       tox -e py26 -- -s tests.integration;
       tox -e py27 -- -s tests.unit &&


.travis.yml
```diff
@@ -1,23 +1,33 @@
...
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_collect.py tests/product/test_server_uninstall.py"
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_connectors.py"
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_control.py"
+
+matrix:
+  include:
+    - python: "2.6"
+      env: OTHER_TESTS=true
+    - python: "2.7"
+      env: OTHER_TESTS=true
```
ArturGajowy: I'd try to run the 'OTHER_TESTS' in a single job, on both python versions using tox. The 'OTHER_TESTS' jobs take 20 min themselves, but should contract to sth less than 2 * 20 min because of sharing the build, image download, etc.

.travis.yml
```diff
@@ -1,26 +1,35 @@
...
-    - SHORT_PRODUCT_TEST_GROUP=0
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN="tests/product/test_server_install.py"
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_status.py"
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_collect.py tests/product/test_server_uninstall.py"
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_connectors.py"
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_control.py"
+
+matrix:
+  include:
+    - python: "2.7"
```
ArturGajowy: now we can get rid of the include: syntax and have python version defined globally. Let's do this before merging.

ArturGajowy:
I'm not sure if I read build output right, but I couldn't find trace of 2.6 tests being executed https://travis-ci.org/prestodb/presto-admin/jobs/160172002

.travis.yml
rschlussel: still experimenting with tox on travis.  This only ran py27.  I was worried if I included both it would create two jobs, but we'll see what happens.  I'll let you know when this is ready to be looked at again.

.travis.yml
```diff
@@ -1,26 +1,29 @@
...
     - LONG_PRODUCT_TESTS="tests/product/test_server_install.py tests/product/test_status.py tests/product/test_collect.py tests/product/test_connectors.py tests/product/test_control.py tests/product/test_server_uninstall.py"
-  matrix:
-    - OTHER_TESTS=true
-    - SHORT_PRODUCT_TEST_GROUP=0
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN="tests/product/test_server_install.py"
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_status.py"
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_collect.py tests/product/test_server_uninstall.py"
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_connectors.py"
-    - LONG_PRODUCT_TEST_GROUP_PRESTO_ADMIN_AND_PRESTO="tests/product/test_control.py"
+
```
ArturGajowy: I think changes in lines 12:22 can be reverted - it's equivalent to the old version, only a bit longer

rschlussel:
Alright cleaned up the commits.  Should be ready for review again.

.travis.yml
rschlussel: done

.travis.yml
rschlussel: done

ArturGajowy:
LGTM! :)

rschlussel:
merged!

PR #257, opened 2016-09-14
Make smoketest and test-rpm run clean-test-all

rschlussel:
@ebd2 small follow on from my previous pr.  Forgot about smoke and test-rpm


rschlussel:
merged

PR #258, opened 2016-09-15
Increase timeout for docker containers to start

rschlussel:
This is to address failures on travis caused by containers not started.

@petroav @ebd2 let's see if this helps before digging further


ebd2:
LGTM

PR #259, opened 2016-09-15
Remove unnecessary test_control tests

rschlussel:
@petroav @cawallin

### DIFF
--- a/tests/product/test_collect.py
+++ b/tests/product/test_collect.py
@@ -67,13 +67,6 @@ class TestCollect(BaseProductTestCase):
         msg = '%s != %s' % (actual, expected)
         return msg
 
-    @attr('smoketest')
-    def test_collect_system_info_basic(self):
-        self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PRESTO_CLUSTER)
-        self.run_prestoadmin('server start')
-        actual = self.run_prestoadmin('collect system_info')
-        self.test_basic_system_info(actual)
-
     @nottest
     def test_basic_system_info(self, actual, coordinator=None, hosts=None):
         if not coordinator:
@@ -132,6 +125,7 @@ class TestCollect(BaseProductTestCase):
                                     [self.cluster.master,
                                      self.cluster.slaves[0]])
 
+    @attr('smoketest')
     def test_system_info_pa_separate_node(self):
         installer = StandalonePrestoInstaller(self)
         self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PA_CLUSTER)
@@ -147,24 +141,6 @@ class TestCollect(BaseProductTestCase):
             hosts=self.cluster.slaves)
 
     @attr('smoketest')
-    def test_collect_query_info(self):
-        self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PRESTO_CLUSTER)
-        self.run_prestoadmin('server start')
-        sql_to_run = 'SELECT * FROM system.runtime.nodes WHERE 1234 = 1234'
-        query_id = self.retry(lambda: self.get_query_id(sql_to_run))
-
-        actual = self.run_prestoadmin('collect query_info ' + query_id)
-        query_info_file_name = path.join(TMP_PRESTO_DEBUG,
-                                         'query_info_' + query_id +
-                                         '.json')
-
-        expected = 'Gathered query information in file: ' + \
-                   query_info_file_name + '\n'
-
-        self.assert_path_exists(self.cluster.master,
-                                query_info_file_name)
-        self.assertEqual(actual, expected)
-
     def test_query_info_pa_separate_node(self):
         installer = StandalonePrestoInstaller(self)
         self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PA_CLUSTER)
--- a/tests/product/test_control.py
+++ b/tests/product/test_control.py
@@ -65,36 +65,9 @@ class TestControl(BaseProductTestCase):
         self.assert_start_with_one_host_started(
             self.cluster.internal_slaves[0])
 
-    def test_server_start_all_started(self):
-        self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PRESTO_CLUSTER)
-        start_output = self.run_prestoadmin('server start').splitlines()
-        process_per_host = self.get_process_per_host(start_output)
-        self.assert_started(process_per_host)
-
-        start_output = self.run_prestoadmin(
-            'server start', raise_error=False).splitlines()
-        self.assertRegexpMatchesLineByLine(
-            start_output,
-            self.expected_port_error(self.cluster.all_internal_hosts())
-        )
-
-    def test_server_stop_not_started(self):
-        self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PRESTO_CLUSTER)
-
-        stop_output = self.run_prestoadmin('server stop').splitlines()
-        not_started_hosts = self.cluster.all_internal_hosts()
-        self.assertRegexpMatchesLineByLine(
-            stop_output,
-            self.expected_stop(not_running=not_started_hosts)
-        )
-
-    def test_server_stop_coordinator_started(self):
+    def test_server_stop_node_started(self):
         self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PRESTO_CLUSTER)
         self.assert_one_host_stopped(self.cluster.internal_master)
-
-    def test_server_stop_worker_started(self):
-        self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PRESTO_CLUSTER)
-        self.assert_one_host_stopped(self.cluster.internal_slaves[0])
 
     def test_server_restart_nothing_started(self):
         self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PRESTO_CLUSTER)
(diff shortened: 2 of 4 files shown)


tests/product/test_control.py
```diff
@@ -65,37 +65,10 @@ def test_server_start_one_host_started(self):
...
-        self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PRESTO_CLUSTER)
-        start_output = self.run_prestoadmin('server start').splitlines()
-        process_per_host = self.get_process_per_host(start_output)
-        self.assert_started(process_per_host)
-
-        start_output = self.run_prestoadmin(
-            'server start', raise_error=False).splitlines()
-        self.assertRegexpMatchesLineByLine(
-            start_output,
-            self.expected_port_error(self.cluster.all_internal_hosts())
```
cawallin: We now have no tests that verify that you can't run start twice (e.g. that you get the "port in use" error)

rschlussel: we have a one host started test

cawallin: ah, missed that. this is fine then.

petroav:
@ebd2 and @anusudarsan can you review this as well? I want us to make sure we're not deleting tests that we added specifically in response to a bug.

tests/product/test_control.py
```diff
@@ -65,37 +65,10 @@ def test_server_start_one_host_started(self):
...
-        )
-
-    def test_server_stop_not_started(self):
```
petroav: This can be removed because `test_server_restart_nothing_started()` covers this scenario (the first step of restart is a stop).

tests/product/test_control.py
```diff
@@ -105,67 +78,24 @@ def test_server_restart_nothing_started(self):
             self.expected_start()[:]
         self.assert_simple_server_restart(expected_output, running_host='')
 
-    def test_server_restart_coordinator_started(self):
```
petroav: The two tests below can be removed because `test_server_stop_node_started()` performs an action when a node is already up.

tests/product/test_control.py
rschlussel: ? it is removed.

tests/product/test_control.py
```diff
@@ -65,37 +65,10 @@ def test_server_start_one_host_started(self):
         self.assert_start_with_one_host_started(
```
petroav: Can you change the test method name in this test (line 63) to match the test method name on line 68? Either both should use "one host" or "node".

tests/product/test_control.py
petroav: Sorry, my bad, I added the comment as an explanation as to why `test_server_stop_not_started()` can be removed.

tests/product/test_control.py
```diff
@@ -65,37 +65,10 @@ def test_server_start_one_host_started(self):
...
-            stop_output,
-            self.expected_stop(not_running=not_started_hosts)
-        )
-
-    def test_server_stop_coordinator_started(self):
+    def test_server_stop_node_started(self):
         self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PRESTO_CLUSTER)
         self.assert_one_host_stopped(self.cluster.internal_master)
 
-    def test_server_stop_worker_started(self):
```
petroav: This can be removed because `test_server_stop_node_started()` covers the case.

tests/product/test_control.py
petroav: And `test_server_start_one_host_started()` covers the other case.

tests/product/test_server_install.py
```diff
@@ -227,24 +227,6 @@ def test_install(self, installer=None):
             self.assert_has_default_config(container)
             self.assert_has_default_connector(container)
 
-    def test_install_worker_is_pa_master(self):
```
cawallin: I think we could also remove test_install because test_install_with_java8_home covers that (and all of our customers set java8_home b/c it's in the docs and they copy paste). test_install_with_java8_home can be in the smoketest

cawallin:
Additional test files look good

tests/product/test_server_install.py
cawallin: also, I'm pretty sure that our normal installer doesn't set java8 home so that path is tested in tests anyway

tests/product/test_control.py
```diff
@@ -186,42 +116,8 @@ def assert_start_stop_restart_coordinator_down(self, coordinator, coordinator_in
...
-        for message in expected_restart:
-            self.assertRegexpMatches(restart_output, message, 'expected %s \n'
-                                                              ' actual %s' % (message, restart_output))
-        restart_output = restart_output.splitlines()
-        process_per_host = self.get_process_per_host(restart_output)
-        self.assert_started(process_per_host)
-
-    def assert_start_stop_restart_down_node(self, down_node,
-                                            down_internal_node):
+    def assert_start_worker_down(self, down_node,
```
petroav: You can put the arguments to the method on a single line.

tests/product/test_server_install.py
rschlussel: ok.  I think it's good to test without java8_home set, but that's covered by all the other tests in this file.

petroav:
ee047c5c6b4515a9bc9524f42893d68158247a9f Looks good except for the test naming. Can you go through and make sure that test methods have consistent naming.

tests/product/test_collect.py
```diff
@@ -67,13 +67,6 @@ def log_msg(self, actual, expected):
         msg = '%s != %s' % (actual, expected)
         return msg
 
-    @attr('smoketest')
```
petroav: This can be removed because `test_system_info_pa_separate_node()` tests a very similar scenario.

tests/product/test_collect.py
```diff
@@ -147,24 +141,6 @@ def test_system_info_pa_separate_node(self):
             hosts=self.cluster.slaves)
 
     @attr('smoketest')
-    def test_collect_query_info(self):
```
petroav: This can be removed because `test_query_info_pa_separate_node()` is basically the same test.

tests/product/test_server_uninstall.py
```diff
@@ -108,54 +78,8 @@ def test_uninstall_lost_host(self):
         cmd_output = self.run_prestoadmin('server uninstall',
                                           raise_error=False)
         self.assertRegexpMatches(cmd_output, expected)
-        process_per_active_host = []
-        for host, pid in process_per_host:
-            if host not in down_node:
-                process_per_active_host.append((host, pid))
-        self.assert_stopped(process_per_active_host)
```
petroav: Why was this being asserted? Does uninstall do a stop before uninstalling?

rschlussel: yes

tests/product/test_server_uninstall.py
```diff
@@ -59,22 +57,6 @@ def assert_uninstalled_dirs_removed(self, container):
         self.assert_path_removed(container, '/usr/shared/doc/presto')
         self.assert_path_removed(container, '/etc/init.d/presto')
 
-    def test_uninstall_when_server_down(self):
```
petroav: This can be removed because `test_uninstall_lost_host()` covers the same scenario.

tests/product/test_server_uninstall.py
```diff
@@ -108,54 +78,8 @@ def test_uninstall_lost_host(self):
...
-            if host not in down_node:
-                process_per_active_host.append((host, pid))
-        self.assert_stopped(process_per_active_host)
 
         for container in [self.cluster.internal_master,
                           self.cluster.internal_slaves[1],
                           self.cluster.internal_slaves[2]]:
             self.assert_uninstalled_dirs_removed(container)
-
-    def test_uninstall_with_dir_readonly(self):
```
petroav: Why can we remove this?

tests/product/test_server_uninstall.py
```diff
@@ -108,54 +78,8 @@ def test_uninstall_lost_host(self):
...
-        uninstalled_hosts = self.cluster.all_hosts()[:]
-        uninstalled_hosts.remove(self.cluster.master)
-
-        for container in uninstalled_hosts:
-            self.assert_uninstalled_dirs_removed(container)
-
-        self.installer.assert_installed(self, container=self.cluster.master)
-
-    @docker_only
-    def test_uninstall_as_non_sudo(self):
```
petroav: Why can we remove this?

petroav:
For f86de533c4630b40798aa6f7d18c71248cdf70e7, can you change the commit message to `Remove tests from test_collect & test_server_uninstall`?

petroav:
Sorry for closing!

tests/product/test_server_uninstall.py
rschlussel: these really test the rpm, not server uninstall.  And the cases aren't useful enough to be worth adding to test_rpm (I could here an argument for uninstall_non_sudo.  The read only directory is just weird.)

tests/product/test_server_install.py
```diff
@@ -407,111 +389,6 @@ def test_install_interactive_with_hostnames(self):
             self.assert_has_default_connector(container)
             self.assert_has_jmx_connector(container)
 
-    def test_install_interactive_with_ips(self):
```
petroav: This can be removed because we're not really testing anything here, just Fabric's ability to connect to a host using an IP.

rschlussel: and we already have test_install_with_ips (not interactive)

rschlussel: it also tests the validation (which is also unit tested, so maybe the other test install with ips isn't necessary), but it's not _just_ a test of fabric

tests/product/test_server_install.py
petroav: This can be removed because it's not really testing presto-admin. It's making sure that there's nothing special about the master host to server as a Presto coordinator.

petroav: I agree, testing without explicitly setting JAVA8_HOME is implicitly tested in all the other tests so we can remove `test_install()`.

tests/product/test_server_install.py
```diff
@@ -407,111 +389,6 @@ def test_install_interactive_with_hostnames(self):
...
-        cmd_output.sort()
-        expected.sort()
-        for expected_regexp, actual_line in zip(expected, cmd_output):
-            self.assertRegexpMatches(actual_line, expected_regexp)
-
-        self.assert_installed_with_regex_configs(
-            self.cluster.master,
-            [self.cluster.slaves[0]])
-
-    def test_install_with_wrong_topology(self):
```
petroav: I assume this is covered by a unit test?

tests/product/test_server_install.py
```diff
@@ -407,111 +389,6 @@ def test_install_interactive_with_hostnames(self):
...
-                   ' host name.' \
-                   '  More detailed information can be found in ' \
-                   '/var/log/prestoadmin/presto-admin.log\n'
-        self.assertRaisesRegexp(OSError,
-                                expected,
-                                self.run_prestoadmin,
-                                'server install /mnt/presto-admin/%(rpm)s ',
-                                rpm=rpm_name)
-
-    def test_install_with_malformed_topology(self):
```
petroav: Unit test?

tests/product/test_server_install.py
```diff
@@ -407,111 +389,6 @@ def test_install_interactive_with_hostnames(self):
...
-                   '  More detailed information can be found in ' \
-                   '/var/log/prestoadmin/presto-admin.log'
-
-        self.assertRaisesRegexp(OSError,
-                                expected,
-                                self.run_prestoadmin,
-                                'server install /mnt/presto-admin/%(rpm)s ',
-                                rpm=rpm_name)
-
-    def test_install_with_malformed_connector(self):
```
petroav: Unit test?

rschlussel:
> For f86de53, can you change the commit message to Remove tests from test_collect & test_server_uninstall?

I was going to squash all the commits together at the end.

tests/product/test_server_install.py
rschlussel: yup

tests/product/test_server_install.py
rschlussel: yup

tests/product/test_server_install.py
rschlussel: yup

tests/product/test_server_uninstall.py
petroav: I vote we keep non-sudo. We'll probably have to change it up a little when we officially support no sudo. Maybe then we can change all tests to run as non-sudo and this one will be sudo?

rschlussel: That's not what it tests.  it tests that the remote user can't uninstall if it doesn't have sudo privileges.  that's going to continue to be true.

tests/product/test_server_uninstall.py
petroav: I assume we don't want to test this in a product test because it's probably already covered in a unit test, i.e. that uninstall calls stop and therefore asserting that everything is stopped is unnecessarily testing the stop action?

rschlussel: we test it in the regular test_server_uninstall also.  and a down node with server stop is tested in test_control.

rschlussel:
merged

PR #260, opened 2016-09-16
Add testing targets for configurable cluster

petroav:
- Those targets don't build the Docker images for testing

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: clean-all clean clean-eggs clean-build clean-pyc clean-test-containers clean-test \
 	clean-docs lint smoke test test-all test-images test-rpm docker-images coverage docs \
 	open-docs release release-builds dist dist-online dist-offline wheel install precommit \
-	clean-test-all
+	clean-test-all smoke-configurable-cluster test-all-configurable-cluster
 
 help:
 	@echo "precommit - run \`quick' tests and tasks that should pass or succeed prior to pushing"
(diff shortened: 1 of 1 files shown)


petroav:
@ebd2

Makefile
```diff
@@ -79,7 +81,11 @@ lint:
 presto-server-rpm.rpm:
 	wget 'https://repository.sonatype.org/service/local/artifact/maven/content?r=central-proxy&g=com.facebook.presto&a=presto-server-rpm&e=rpm&v=RELEASE' -O $@
 
-smoke: clean-test-all test-images
+smoke: clean-test-all test-images _smoke
+
+smoke-configurable-cluster: clean-test-all _smoke
```
rschlussel: It probably shouldn't clean the test images either

Makefile
```diff
@@ -88,7 +94,11 @@ test: clean-test
 
 TEST_SUITE?=tests.product
 
-test-all: clean-test-all test-images
+test-all: clean-test-all test-images _test-all
+
+test-all-configurable-cluster: clean-test-all _test-all
```
rschlussel: ditto

Makefile
petroav: Hmm, I added that because I wanted to make sure the contents of the tmp directory were cleared. That directory contains the offline installed used with Docker but with a configurable cluster I can only use the online installer.

Maybe I can do a separate target for cleaning that directory?

petroav: Or actually I can move the `rm -rf tmp` to `clean-test`. I think that should be better than introducing yet another clean target. Thoughts?

rschlussel: I care more that `clean-test` only cleans the things from `make test` than that this doesn't clean more than it needs to (I sometimes run `make test` while in the middle of debugging a product test).   I'm not bothered by adding another clean target, since it matches the extra test targets, but if you want to leave this as is, that's fine too.

PR #262, opened 2016-09-19
Collect connectors

rschlussel:


akshatnair:
LGTM but what about docs? https://github.com/prestodb/presto-admin/blob/master/docs/presto-admin-commands.rst

rschlussel:
The docs already say they get info about connectors configured, so I thought this was just more info. Would you rather I explicitly say that it gathers connector configuration files?

akshatnair:
my bad I thought this was a new command

akshatnair:
but that I think meant the name from server status? but now we also collect the config details so maybe add a line connectors configurations rather than connectors configured?

rschlussel:
done (squashed it into the last commit)

rschlussel:
merged

PR #263, opened 2016-09-19
Use a shell when appending to node.properties

cawallin:
For the sake of consistency, set files.append to use a shell like all the
other presto-admin commands.


cawallin:
This is applicable to the Hadoop appliance docs, which recommend giving limited sudo privileges to the prestoadmin user (i.e. only /bin/bash)

sanjay990:
LGTM

PR #265, opened 2016-09-21
Remove unnecessary tests

rschlussel:
Tests removed when the functionality is sufficiently unit tested, or
there is another product test covering the case.

I removed more tests!  hopefully the prestodb/presto-admin builds will stop timing out on connector tests and short product tests.   @petroav @ebd2


tests/product/test_configuration.py
```diff
@@ -221,23 +218,6 @@ def test_lost_coordinator_connection(self):
         self.assertRegexpMatches(str.join('\n', output.splitlines()[6:]),
                                  expected)
 
-    def test_deploy_lost_worker_connection(self):
```
rschlussel: there's already a test for a lost coordinator connection and we don't need both.

tests/product/test_connectors.py
```diff
@@ -118,66 +120,6 @@ def test_connector_add_remove_coord_worker_using_dash_x(self):
         for slave in [self.cluster.slaves[1], self.cluster.slaves[2]]:
             self.assert_has_default_connector(slave)
 
-    @docker_only
-    def test_connector_add_wrong_permissions(self):
```
rschlussel: testing the file system

tests/product/test_connectors.py
```diff
@@ -118,66 +120,6 @@ def test_connector_add_remove_coord_worker_using_dash_x(self):
...
-        self.setup_cluster(NoHadoopBareImageProvider(), STANDALONE_PRESTO_CLUSTER)
-
-        # test add a connector that does not exist
-        not_found_error = self.fatal_error(
-            'Configuration for connector tpch not found')
-        self.run_prestoadmin('connector remove tpch')
-        self.assertRaisesRegexp(OSError, not_found_error,
-                                self.run_prestoadmin, 'connector add tpch')
-
-    def test_connector_add_no_dir(self):
```
rschlussel: unit tested

tests/product/test_connectors.py
```diff
@@ -118,66 +120,6 @@ def test_connector_add_remove_coord_worker_using_dash_x(self):
...
-
-        # test add connector by file without read permissions on directory
-        script = 'chmod 600 /etc/opt/prestoadmin/connectors; ' \
-                 'su app-admin -c "./presto-admin connector add tpch"'
-        not_found_error = self.fatal_error(
-            'Configuration for connector tpch not found')
-        self.assertRaisesRegexp(OSError, not_found_error,
-                                self.run_script_from_prestoadmin_dir, script)
-
-    def test_connector_add_missing_connector(self):
```
rschlussel: unit tested

tests/product/test_connectors.py
```diff
@@ -217,31 +159,6 @@ def test_connector_add_empty_dir(self):
 """
         self.assertEqualIgnoringOrder(expected, output)
 
-    def test_connector_add_two_connectors(self):
```
rschlussel: changed the basic smoketest to add two connectors

tests/product/test_connectors.py
```diff
@@ -353,22 +270,6 @@ def test_connector_remove(self):
             self.run_prestoadmin,
             'connector remove tpch')
 
-    def test_connector_name_not_found(self):
```
rschlussel: unit tested

tests/product/test_package_install.py
```diff
@@ -91,91 +91,6 @@ def test_install_exclude_nodes(self):
...
-        error = '\nFatal error: [%s] error: ' \
-                '/mnt/presto-admin/invalid-path/presto.rpm: open failed: ' \
-                'No such file or directory\n\nAborting.\n'
-        expected = ''
-        for host in self.cluster.all_internal_hosts():
-            expected += error % host
-
-        self.assertEqualIgnoringOrder(cmd_output, expected)
-
-    def test_install_no_path_arg(self):
```
rschlussel: tested by the error_handling product tests

tests/product/test_package_install.py
```diff
@@ -91,91 +91,6 @@ def test_install_exclude_nodes(self):
         for container in self.cluster.all_hosts():
             self.installer.assert_uninstalled(container, msg=output)
 
-    def test_install_invalid_path(self):
```
rschlussel: unit tested

tests/product/test_package_install.py
```diff
@@ -91,91 +91,6 @@ def test_install_exclude_nodes(self):
...
-                                 '        local_path: Absolute path to the rpm'
-                                 ' to be installed\n        '
-                                 '--nodeps (optional): Flag to indicate if '
-                                 'rpm install\n'
-                                 '            should ignore c'
-                                 'hecking package dependencies. Equivalent\n'
-                                 '            to adding --nodeps flag to rpm '
-                                 '-i.\n\n')
-
-    def test_install_already_uninstalled(self):
```
rschlussel: we already unit test that it calls to rpm.  So this is just testing rpm

tests/product/test_package_install.py
```diff
@@ -91,91 +91,6 @@ def test_install_exclude_nodes(self):
...
-        output = self.run_prestoadmin('package uninstall some_rpm -H %(master)s', raise_error=False)
-        expected = self.escape_for_regex(self.replace_keywords(
-            '\n\nAborting.\nFatal error: [%(master)s] Package is not installed',
-            **self.installer.get_keywords()))
-        self.assertRegexpMatchesLineByLine(
-            output.splitlines(),
-            expected.splitlines()
-        )
-
-    def test_install_already_installed(self):
```
rschlussel: ditto

tests/product/test_topology.py
```diff
@@ -58,26 +58,6 @@ def test_topology_show(self):
         expected = normal_topology
         self.assertEqual(expected, actual)
 
-    @docker_only
```
rschlussel: topology show doesn't connect to any remote hosts, so this is a weird thing to test.

tests/product/test_configuration.py
```diff
@@ -337,16 +317,7 @@ def test_configuration_show_coord_worker_using_dash_x(self):
             expected = f.read()
         self.assertRegexpMatches(output, expected)
 
-    def test_configuration_no_presto_user(self):
```
rschlussel: i don't remember why i deleted this one.  Maybe I should add it back?

tests/product/test_configuration.py
petroav: The other product test that tests this is `test_lost_coordinator_connection`.

tests/product/test_configuration.py
petroav: There is a test in `test_connectors` (`test_connector_add_no_presto_user`) that does the same thing but for adding a connector. I say we can get rid of this one and add a unit test that ensures that we try to add configuration files as the presto user.

tests/product/test_connectors.py
```diff
@@ -43,30 +42,33 @@ def setup_cluster_assert_connectors(self):
...
+    def test_connector_add_remove(self):
         self.setup_cluster_assert_connectors()
-
         self.run_prestoadmin('connector remove tpch')
-        self.run_prestoadmin('server restart')
         self.assert_path_removed(self.cluster.master,
                                  os.path.join(constants.CONNECTORS_DIR,
                                               'tpch.properties'))
-        self._assert_connectors_loaded([['system']])
-        for host in self.cluster.all_hosts():
```
petroav: I say we keep this for loop, it's not that expensive. I agree that we should remove the `server restart` and the `_assert_connectors_loaded` because those test that Presto picks up connectors. But checking that the file was removed on the file system on all nodes should be part of this test.

tests/product/test_connectors.py
```diff
@@ -43,30 +42,33 @@ def setup_cluster_assert_connectors(self):
...
             'connector.name=tpch',
             os.path.join(constants.CONNECTORS_DIR, 'tpch.properties'),
             self.cluster.master
         )
+        self.cluster.write_content_to_host(
+            'connector.name=jmx',
+            os.path.join(constants.CONNECTORS_DIR, 'jmx.properties'),
+            self.cluster.master
+        )
         self.run_prestoadmin('connector add')
```
petroav: How does it know what connector to add? Or does it default to adding everything inside its connector directory?

tests/product/test_package_install.py
petroav: The fact that those tests are using a separate command doesn't matter? I assume the same logic is exercised?

tests/product/test_package_install.py
```diff
@@ -91,91 +91,6 @@ def test_install_exclude_nodes(self):
...
-Deploying rpm on %(master)s...
-Package deployed successfully on: %(master)s
-[%(master)s] out: 	package %(rpm_basename)s is already installed
-[%(master)s] out: """, **self.installer.get_keywords()))
-        self.assertRegexpMatchesLineByLine(
-            cmd_output.splitlines(),
-            expected.splitlines()
-        )
-
-    def test_install_not_an_rpm(self):
```
petroav: This is also testing the rpm: if we give something other than an rpm file to the rpm utility, we make sure it crashes.

tests/product/test_topology.py
petroav: Can we unit test this? I assume it's just reading the config.json locally and returning that.

tests/unit/test_package.py
```diff
@@ -119,6 +119,12 @@ def test_deploy_with_fallback_location(self, mock_put, mock_sudo, mock_isfile):
                                     use_sudo=True,
                                     temp_dir='/tmp')
 
+    @patch('prestoadmin.package.os.path.isfile')
+    def test_deploy_invalid_local_path(self, mock_isfile):
+        mock_isfile.return_value = False
+        invalid_path = '/invalid/path'
+        self.assertRaisesRegexp(SystemExit, 'RPM file not found at %s' % invalid_path,  package.deploy, invalid_path)
```
petroav: It looks like you have an extra space after the command and the `package.deploy` function.

tests/product/test_package_install.py
rschlussel: yup.

tests/product/test_connectors.py
rschlussel: adds all connectors in the directory

tests/product/test_connectors.py
rschlussel: oh yes,  didn't mean to remove

tests/product/test_configuration.py
rschlussel: I actually think this is valuable, as it makes sure that our files end up with the right permissions.  I'm going to add this test back.  I'll add some unit tests when I fix #266.  Maybe we can revisit then and see if those are sufficient.

tests/unit/test_package.py
rschlussel: fixed

tests/product/test_topology.py
petroav: No need to do this, after in person discussion.

rschlussel:
merged

PR #269, opened 2016-09-23
Amend product tests to pass on EMR

petroav:


petroav:
@ebd2 & @rschlussel

tests/product/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import contextlib
+import os
+
+
+def determine_jdk_directory(cluster):
```
rschlussel: what if there are two directories that match?

tests/product/base_product_case.py
```diff
@@ -92,12 +92,20 @@ class BaseProductTestCase(BaseTestCase):
                                   r'|\nWarning: (\[%(host)s] )?Timed out ' \
                                   r'trying to connect to %(host)s \(tried 1 ' \
                                   r'time\)\n\nUnderlying exception:' \
-                                  r'\n    timed out\n)'
+                                  r'\n    timed out\n)' \
```
rschlussel: Can you add a comment at the top of these regexes to explain what they are, since they're a bit hard to read if you don't already know.

tests/product/test_server_install.py
```diff
@@ -445,10 +441,21 @@ def test_install_non_root_user(self):
...
+                rpm_dir=self.cluster.get_rpm_cache_dir(), name=rpm_name)
         )
 
         for container in self.cluster.all_hosts():
             installer.assert_installed(self, container)
             self.assert_has_default_config(container)
             self.assert_has_default_connector(container)
+
+    @staticmethod
+    def format_err_msgs_with_internal_hosts(cluster, msgs):
```
rschlussel: why make this a static method and pass in the cluster instead of just calling self.cluster.x?

It would also make the method call shorter, which would be easier to read.

tests/product/__init__.py
petroav: It'll get back all directories that match, it will then separate all components of all those directories and pick the last one as the JDK to use.

Hopefully this shouldn't be a problem since we control the configurable cluster and docker environments so we know what JDKs are installed. I would do a check in this method for the number of matches and if there are more than 1 then I can fail fast. I think that's a good solution. What do you think?

tests/product/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+    Return the directory where the JDK is installed. For example if the JDK is
+    located in /usr/java/jdk1.8_91, then this method will return the string
+    'jdk1.8_91'
+
+    :param cluster: cluster object on which to determine the JDK directory
+    """
+    output = cluster.exec_cmd_on_host(cluster.master, 'bash -c "ls -d /usr/java/jdk*"')
```
ebd2: Nice use of ls -d

tests/product/__init__.py
ebd2: I'm in agreement that failing fast sounds appropriate. If the env is screwed up, we don't want to spend several hours waiting for basically invalid test results.

tests/product/__init__.py
```diff
@@ -0,0 +1,52 @@
...
+    return output.split(os.path.sep)[-1][:-1]  # trim newline
```
ebd2: use string.strip

tests/product/__init__.py
petroav: I learn from the best :)

tests/product/__init__.py
petroav: I just did a refactor so that 1 is possible but we might want to do 2. I don't feel strongly about either. If there's more than one JDK then most tests will pass if at least one JDK is correctly configured.

tests/product/base_product_case.py
petroav: Done

tests/product/test_server_install.py
petroav: No good reason really. I was expecting to re-use the method in other classes. I'll make it an instance method.

tests/product/__init__.py
petroav: Done. Comment removed since code makes it clear what I'm doing.

PR #270, opened 2016-09-23
Permissions of deployed files and keep node.id

### LINKED ISSUE #266: Configuration deploy doesn't keep the node.id
When you run `configuration deploy`, the `node.properties` file gets entirely overwritten.  It should not overwrite the `node.id`.

This happens in deploy.deploy_node_properties() at the call to secure_create_file().

rschlussel:
fixes #266

### DIFF
--- a/prestoadmin/connector.py
+++ b/prestoadmin/connector.py
@@ -24,6 +24,7 @@ from fabric.context_managers import hide
 from fabric.contrib import files
 from fabric.operations import sudo, os, get
 
+from prestoadmin.deploy import secure_create_directory
 from prestoadmin.standalone.config import StandaloneConfig, \
     PRESTO_STANDALONE_USER_GROUP
 from prestoadmin.util import constants
@@ -42,7 +43,7 @@ COULD_NOT_REMOVE = 'Could not remove connector'
 
 def deploy_files(filenames, local_dir, remote_dir, user_group, mode=0600):
     _LOGGER.info('Deploying configurations for ' + str(filenames))
-    sudo('mkdir -p ' + remote_dir)
+    secure_create_directory(remote_dir, PRESTO_STANDALONE_USER_GROUP)
     for name in filenames:
         put_secure(user_group, mode, os.path.join(local_dir, name), remote_dir,
                    use_sudo=True)
--- a/prestoadmin/connector.py
+++ b/prestoadmin/connector.py
@@ -41,6 +41,8 @@ __all__ = ['add', 'remove']
 COULD_NOT_REMOVE = 'Could not remove connector'
 
 
+# we deploy connector files with 0600 permissions because they can contain passwords
+# that should not be world readable
 def deploy_files(filenames, local_dir, remote_dir, user_group, mode=0600):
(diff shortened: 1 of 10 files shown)


prestoadmin/deploy.py
```diff
@@ -142,7 +143,11 @@ def deploy_node_properties(content, remote_dir):
     _LOGGER.info("Deploying node.properties configuration")
     name = "node.properties"
     node_file_path = (os.path.join(remote_dir, name))
-    secure_create_file(node_file_path, PRESTO_STANDALONE_USER_GROUP)
+    if not exists(node_file_path, use_sudo=True):
+        secure_create_file(node_file_path, PRESTO_STANDALONE_USER_GROUP, mode=644)
+    else:
```
rschlussel: we need to do this because the rpm doesn't set up the config files with correct ownership (see https://github.com/Teradata/presto/pull/358)

rschlussel:
@petroav @ebd2

tests/product/base_product_case.py
```diff
@@ -267,8 +267,19 @@ def get_file_content(self, host, filepath):
 
     def assert_config_perms(self, host, filepath):
         self.assert_file_perm_owner(
+            host, filepath, '-rw-r--r--', 'presto', 'presto')
+
+    def assert_connnector_perms(self, host, filepath):
```
ebd2: typo: connnector

tests/product/base_product_case.py
```diff
@@ -267,8 +267,19 @@ def get_file_content(self, host, filepath):
...
+        self.assert_file_perm_owner(
             host, filepath, '-rw-------', 'presto', 'presto')
 
+    def assert_directory_perm_owner(self, host, filepath, permissions, owner, group):
+        ls = self.cluster.exec_cmd_on_host(host, "ls -l -d %s" % (filepath,))
```
ebd2: Everything but the ls command is the same in this and assert_file_perm_owner. Please consolidate.

tests/product/base_product_case.py
```diff
@@ -267,8 +267,19 @@ def get_file_content(self, host, filepath):
...
+        fields = ls.split()
+        self.assertEqual(fields[0], permissions)
```
ebd2: we should probably assert that the expected perms includes the 'd' or 'l' at the beginning or there's no way the perms are going to match. This assertion will fail, but it's misleading because it's the expected value that's totally bogus.

ebd2: Actually, let's not worry about symlinks for now...

tests/product/base_product_case.py
rschlussel: done

prestoadmin/connector.py
```diff
@@ -41,7 +42,7 @@
 
 def deploy_files(filenames, local_dir, remote_dir, user_group, mode=0600):
```
rschlussel: done

tests/product/base_product_case.py
rschlussel: done

PR #273, opened 2016-09-28
Pin docker images to version 6

ebd2:
We've been pulling latest, which was fine until we updated the latest tag on dockerhub with a new version of java. Tests that rely on java being where they expect it to be now fail. This PR pins the docker images to a version that has java in the expected location.

### DIFF
--- a/docs/release/release-1.4.2.rst
+++ b/docs/release/release-1.4.2.rst
@@ -1,0 +1,7 @@
+=============
+Release 1.4.1
+=============
+
+This release works for Presto versions 0.116-0.148.
+
+* Pin docker images used for product tests to version 6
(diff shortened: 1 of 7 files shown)


rschlussel:
looks good

docs/release/release-1.4.2.rst
```diff
@@ -0,0 +1,7 @@
+=============
+Release 1.4.1
```
petroav: Wrong release number, should be 1.4.2.

petroav:
@ebd2 I think this was merged so you can close it. Also don't forget to delete your branch on our fork if you haven't already done so.

ebd2:
Hmmm. I wonder why this didn't close automatically when I pushed. I doubt that I had to rebase, after all. Thanks for catching that @petroav

PR #276, opened 2016-09-28
Convert variables to read only @properties

petroav:


petroav:
@rschlussel & @ebd2 
Passing Travis build https://travis-ci.org/Teradata/presto-admin/builds/163501390

rschlussel:
what was the motivation for this?  If you want to make these fields common to the abstract class, you should use `@abstractproperty`.  see http://stackoverflow.com/questions/5960337/how-to-create-abstract-properties-in-python-abstract-classes  or https://docs.python.org/2/library/abc.html

petroav:
The motivation for this is that I wanted to introduce some kind of immutability to the code and to clean it up a little.

I'll further re-factor so that the properties are in the abstract class.

petroav:
@ebd2 do you want to take another look? I updated the commit. You probably only want to look at `base_cluster.py` because everything else is the same.

ebd2:
LGTM

PR #279, opened 2016-10-10
Add G1HeapRegionSize property to the JVM defaults

arhimondr:
As it is suggested in Presto documentation https://prestodb.io/docs/current/installation/deployment.html. Omitting this property might potentially cause GC issues.


cawallin:
You should check out the unit/product tests, they're going to fail unless you update these properties in the results

cawallin:
thanks for fixing this though :)

arhimondr:
@cawallin I'm going to wait for the tests to fail, and than i will fix all the failures at once =)

PR #280, opened 2016-10-13
Switch plugin.config-dir to catalog.config-dir

cawallin:
The Presto server logs have the following warning: "Warning: Configuration
property 'plugin.config-dir' has been replaced. Use 'catalog.config-dir' instead."
Keep plugin.config-dir as '/etc/presto', since that's also a valid Presto config with
the default 'etc'.


rschlussel:
looks good % passing tests.  Thanks for fixing this!

PR #282, opened 2016-10-24
Add X-Presto-Source to prestoclient headers

cawallin:
Otherwise, the web UI in 0.152 doesn't load, because there's a bug
with null headers that's fixed in 0.153.

### DIFF
--- a/prestoadmin/prestoclient.py
+++ b/prestoadmin/prestoclient.py
@@ -88,7 +88,8 @@ class PrestoClient:
 
         headers = {"X-Presto-Catalog": catalog,
                    "X-Presto-Schema": schema,
-                   "X-Presto-User": self.user}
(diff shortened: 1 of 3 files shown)


ebd2:
Have you tested this with both old and new versions of Presto?

cawallin:
I've tested this with 152t and it fixes the issue, I will try with 154.

ebd2:
What about 148t? Is this backwards compatible?

cawallin:
The source header should be sent on every request; the lack of doing so is a little bit of a bug. So it should be backwards compatible, I can check if you think it'd be helpful (within reason, we don't need to check too many versions)

ebd2:
LGTM then.

cawallin:
I tried 152t, 148t, and 154, and all of them work.

PR #288, opened 2016-11-30
Https client

ebd2:
Super rough for initial feedback.

### DIFF
--- a/packaging/install-prestoadmin.template
+++ b/packaging/install-prestoadmin.template
@@ -37,7 +37,7 @@ if [ -d "third-party" ]; then
     tar xvzf third-party/virtualenv-%VIRTUALENV_VERSION%.tar.gz -C third-party || true
     "$PYTHON_BIN" third-party/virtualenv-%VIRTUALENV_VERSION%/virtualenv.py presto-admin-install
 else
-    wget https://pypi.python.org/packages/source/v/virtualenv/virtualenv-%VIRTUALENV_VERSION%.tar.gz
+    wget --no-check-certificate https://pypi.python.org/packages/source/v/virtualenv/virtualenv-%VIRTUALENV_VERSION%.tar.gz
     tar xvzf virtualenv-%VIRTUALENV_VERSION%.tar.gz || true
     "$PYTHON_BIN" virtualenv-%VIRTUALENV_VERSION%/virtualenv.py presto-admin-install
 fi
--- a/prestoadmin/configure_cmds.py
+++ b/prestoadmin/configure_cmds.py
@@ -19,26 +19,23 @@ import logging
 import os
 from StringIO import StringIO
 from contextlib import closing
+
+import prestoadmin.deploy
 from fabric.contrib import files
 from fabric.decorators import task, serial
 from fabric.operations import get
 from fabric.state import env
 from fabric.utils import abort, warn
-import prestoadmin.deploy
 from prestoadmin.standalone.config import StandaloneConfig, \
     PRESTO_STANDALONE_USER_GROUP
 from prestoadmin.util import constants
 from prestoadmin.util.base_config import requires_config
+from prestoadmin.util.constants import CONFIG_PROPERTIES, LOG_PROPERTIES, \
+    JVM_CONFIG, NODE_PROPERTIES
 from prestoadmin.util.fabricapi import put_secure
 from prestoadmin.util.filesystem import ensure_parent_directories_exist
 
 __all__ = ['show']
-
-
-CONFIG_PROPERTIES = "config.properties"
-LOG_PROPERTIES = "log.properties"
-JVM_CONFIG = "jvm.config"
-NODE_PROPERTIES = "node.properties"
 
 ALL_CONFIG = [CONFIG_PROPERTIES, LOG_PROPERTIES, JVM_CONFIG, NODE_PROPERTIES]
 
--- a/prestoadmin/prestoclient.py
+++ b/prestoadmin/prestoclient.py
@@ -15,17 +15,23 @@
 """
 Simple client to communicate with a Presto server.
 """
-from httplib import HTTPConnection, HTTPException
-import logging
 import json
+import logging
+import os
 import socket
+import urlparse
+from httplib import HTTPConnection, HTTPException
+from tempfile import mkstemp
 
-from urllib2 import HTTPError, urlopen, URLError
-
-from prestoadmin.util.remote_config_util import lookup_port
-
+from StringIO import StringIO
+from fabric.operations import get
+from fabric.utils import error
+from fabric.state import env
+from jks import jks, base64, textwrap
 from prestoadmin.util.exception import InvalidArgumentError
-
+from prestoadmin.util.httpscacertconnection import HTTPSCaCertConnection
+from prestoadmin.util.local_config_util import get_coordinator_directory, get_topology_path
+from prestoadmin.util.presto_config import PrestoConfig
 
 _LOGGER = logging.getLogger(__name__)
 URL_TIMEOUT_MS = 5000
@@ -33,17 +39,33 @@ NUM_ROWS = 1000
 DATA_RESP = "data"
 NEXT_URI_RESP = "nextUri"
 
+CERTIFICATE_ALIAS = 'certificate_alias'
+
 
 class PrestoClient:
-    response_from_server = {}
-    # rows returned by the query
-    rows = []
-    next_uri = ""
-
-    def __init__(self, server, user, port=None):
+    def __init__(self, server, user):
+        # immutable stuff
         self.server = server
         self.user = user
-        self.port = port
+        self.coordinator_config = PrestoConfig.coordinator_config()
+        self.port = PrestoClient._get_configured_port(self.coordinator_config)
+
+        # mutable stuff
+        self.ca_file_path = ""
+        self.keystore_data = ""
+        self.rows = []
+        self.next_uri = ''
+        self.response_from_server = {}
+
+    @staticmethod
+    def _remove_silently(path):
+        try:
+            os.remove(path)
+        except:
+            pass
+
+    def close(self):
+        PrestoClient._remove_silently(self.ca_file_path)
 
     def clear_old_results(self):
         if self.rows:
@@ -81,9 +103,6 @@ class PrestoClient:
         if not self.user:
             raise InvalidArgumentError("Username missing")
 
-        if not self.port:
-            self.port = lookup_port(self.server)
-
         self.clear_old_results()
 
         headers = {"X-Presto-Catalog": catalog,
@@ -95,8 +114,7 @@ class PrestoClient:
             _LOGGER.info("Connecting to server at: " + self.server +
                          ":" + str(self.port) + " as user " + self.user +
                          " to execute query " + sql)
-            conn = HTTPConnection(self.server, self.port, False,
-                                  URL_TIMEOUT_MS)
+            conn = self._get_connection()
             conn.request("POST", "/v1/statement", sql, headers)
             response = conn.getresponse()
 
@@ -126,18 +144,32 @@ class PrestoClient:
         Sends a GET request to the Presto server at the specified next_uri
         and updates the response
         """
-        try:
-            conn = urlopen(uri, None, URL_TIMEOUT_MS)
-            answer = conn.read()
-            conn.close()
 
-            self.response_from_server = json.loads(answer)
-            _LOGGER.info("GET request successful for uri: " + uri)
-            return True
-        except (HTTPError, URLError) as e:
-            _LOGGER.error("Error opening the presto response uri: " +
-                          str(e.reason))
+        """
+        Remove the scheme and host/port from the uri; the connection itself
+        has that information.
+        """
+        parts = list(urlparse.urlsplit(uri))
+        parts[0] = None
+        parts[1] = None
+        location = urlparse.urlunsplit(parts)
+
+        conn = self._get_connection()
+        conn.request("GET", location)
+        response = conn.getresponse()
+
+        if response.status != 200:
+            conn.close()
+            _LOGGER.error("Error making GET request to %s: %s %s" %
+                          (uri, response.status, response.reason))
             return False
+
+        answer = response.read()
+        conn.close()
+
+        self.response_from_server = json.loads(answer)
+        _LOGGER.info("GET request successful for uri: " + uri)
+        return True
 
     def build_results_from_response(self):
         """
@@ -191,3 +223,90 @@ class PrestoClient:
 
     def get_next_uri(self):
         return self.next_uri
+
+    def _get_connection(self):
+        if self.coordinator_config.use_https():
+            return self._get_https_connection()
+        else:
+            return HTTPConnection(self.server, self.port, False, URL_TIMEOUT_MS)
+
+    @staticmethod
+    def _get_configured_port(coordinator_config):
+        if coordinator_config.use_https():
+            return coordinator_config.get_https_port()
+        else:
+            return coordinator_config.get_http_port()
+
+    def _get_https_connection(self):
+        ca_file_path = self._get_pem()
+        result = HTTPSCaCertConnection(
+                self.server, self.port, None, None, ca_file_path, False, URL_TIMEOUT_MS)
+        return result
+
+    def _fetch_keystore_data(self):
+        if not self.keystore_data:
+            remote_keystore_path = self.coordinator_config.get_client_keystore_path()
+            keystore_data = StringIO()
+            get(remote_keystore_path, keystore_data, use_sudo=True)
+            keystore_data.seek(0)
+            self.keystore_data = keystore_data.getvalue()
+        return self.keystore_data
+
+    def _pem_string(self, der_bytes, type):
+        result = "-----BEGIN %s-----\n" % type
+        result += "\r\n".join(
+            textwrap.wrap(base64.b64encode(der_bytes).decode('ascii'), 64))
+        result += "\n-----END %s-----\n" % type
+        return result
+
+    def _write_pem_file(self, directory, der_bytes_list, type):
+        prefix = os.path.join(directory,
+                              '%s-' % type.lower().replace(' ', '-'))
+        fd, pem_path = mkstemp('.pem', prefix)
+        # https://www.digicert.com/ssl-support/pem-ssl-creation.htm
+        with open(pem_path, 'w') as pem_file:
+            for der_bytes in der_bytes_list:
+                pem_file.write(self._pem_string(der_bytes, type))
+        os.close(fd)
+        return pem_path
+
+    def _get_pem(self):
+        keystore_data = self._fetch_keystore_data()
+
+        keystore = jks.KeyStore.loads(
+                keystore_data,
+                self.coordinator_config.get_client_keystore_password())
+
+        if len(keystore.private_keys.items()) == 1:
+            _, private_key = keystore.private_keys.items()[0]
+        else:
+            private_key = self._get_private_key(keystore)
+        if not self.ca_file_path:
+            """
+            Each member of the cert chain is a tuple (cert_type, cert_data)
+            We only need to write the data out to the .PEM file.
+
+            This usage is shown in the example in the README.md on github:
+            https://github.com/kurtbrose/pyjks
+            """
+            self.ca_file_path = self._write_pem_file(
+                    get_coordinator_directory(),
+                    [cert[1] for cert in private_key.cert_chain], 'CERTIFICATE')
+
+        return self.ca_file_path
+
+    def _get_private_key(self, keystore):
+        all_keys = ", ".join(keystore.private_keys.keys())
+        try:
+            alias = env.conf[CERTIFICATE_ALIAS]
+        except KeyError:
+            error('Multiple keys found in %s. Set %s in %s. Available aliases are %s' %
+                  (self.coordinator_config.get_client_keystore_path(),
+                   CERTIFICATE_ALIAS, get_topology_path(), all_keys))
+
+        try:
+            return keystore.private_keys[alias]
+        except KeyError:
+            error('No alias %s found in %s. Available aliases are %s' %
+                  (alias, self.coordinator_config.get_client_keystore_path(),
+                   all_keys))
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -22,6 +22,7 @@ import re
 import sys
 import urllib2
 import urlparse
+from contextlib import closing
 from tempfile import mkdtemp
 
 from fabric.api import task, sudo, env
@@ -621,13 +622,13 @@ def check_server_status():
     """
     if len(get_coordinator_role()) < 1:
         warn('No coordinator defined.  Cannot verify server status.')
-    client = PrestoClient(get_coordinator_role()[0], env.user)
-    node_id = lookup_string_config('node.id', os.path.join(constants.REMOTE_CONF_DIR, 'node.properties'), env.host)
+    with closing(PrestoClient(get_coordinator_role()[0], env.user)) as client:
+        node_id = lookup_string_config('node.id', os.path.join(constants.REMOTE_CONF_DIR, 'node.properties'), env.host)
 
-    try:
-        return query_server_for_status(client, node_id)
-    except RetryError:
-        return False
+        try:
+            return query_server_for_status(client, node_id)
+        except RetryError:
+            return False
 
 
 @retry(stop_max_delay=RETRY_TIMEOUT * 1000, wait_fixed=5000, retry_on_result=lambda result: result is False)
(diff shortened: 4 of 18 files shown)


packaging/install-prestoadmin.template
```diff
@@ -37,7 +37,7 @@ if [ -d "third-party" ]; then
     tar xvzf third-party/virtualenv-%VIRTUALENV_VERSION%.tar.gz -C third-party || true
     "$PYTHON_BIN" third-party/virtualenv-%VIRTUALENV_VERSION%/virtualenv.py presto-admin-install
 else
-    wget https://pypi.python.org/packages/source/v/virtualenv/virtualenv-%VIRTUALENV_VERSION%.tar.gz
+    wget --no-check-certificate https://pypi.python.org/packages/source/v/virtualenv/virtualenv-%VIRTUALENV_VERSION%.tar.gz
```
ebd2: This needs to be reverted; for testing only.

prestoadmin/prestoclient.py
```diff
@@ -191,3 +219,91 @@ def get_rows(self, num_of_rows=NUM_ROWS):
...
+                    get_coordinator_directory(),
+                    [cert[1] for cert in private_key.cert_chain], 'CERTIFICATE')
+
+        return self.ca_file_path
+
+    def _get_private_key(self, keystore, pa_config=None):
+        try:
+            alias = pa_config[CERTIFICATE_ALIAS]
+            return keystore.private_keys[alias]
+        except KeyError:
```
ebd2: This won't work at execution time because it requires prompting when we're probably in parallel mode. That actually shouldn't occur unless the keystore gets replaced with one that doesn't contain the alias between config and execution, so I'm not overly worried about this. Thoughts?

prestoadmin/prestoclient.py
```diff
@@ -191,3 +219,91 @@ def get_rows(self, num_of_rows=NUM_ROWS):
...
+
+class AliasConfigItem(object):
+    def __init__(self):
+        self.key = CERTIFICATE_ALIAS
+
+    def prompt_user(self, conf):
+        with closing(PrestoClient(None, None)) as client:
+            if not client.requires_https():
+                return
+            client.get_pem(conf)
```
ebd2: Maybe move the KeyError catch here to avoid the runtime prompt error, which is misleading?

prestoadmin/server.py
```diff
@@ -772,69 +773,69 @@ def print_status_header(external_ip, server_status, host):
...
-        coordinator_status = []
-        catalog_status = []
-
-    with settings(hide('running')):
-        node_information = execute(collect_node_information,
-                                   hosts=get_host_list())
-
-    for host in get_host_list():
-        if isinstance(node_information[host], Exception):
+    with closing(PrestoClient(get_coordinator_role()[0], env.user)) as client:
```
ebd2: This looks like collateral damage from resolving conflicts. I'll look into it.

prestoadmin/prestoclient.py
```diff
@@ -15,35 +15,55 @@
...
-    rows = []
-    next_uri = ""
 
+class PrestoClient:
     def __init__(self, server, user, port=None):
         self.server = server
         self.user = user
         self.port = port
+        self.coordinator_config = PrestoConfig.coordinator_config()
+        self.ca_file_path = ""
```
cawallin: can you assign this up front?

prestoadmin/server.py
ebd2: Actually, I think this is just a super misleading diff caused by the whitespace change. Please look extra-close :-(

prestoadmin/util/base_config.py
```diff
@@ -29,6 +29,44 @@
...
+    prompt_string += "Please enter a number between 0 and %s" % (i,)
+
+    answer = prompt(
+            prompt_string,
+            validate=lambda candidate: validate_integer(candidate, 0, i))
+
+    return choices[answer]
+
+
+class SetEnvConfigItem(object):
```
ebd2: hack, hack, hack ;-)

prestoadmin/prestoclient.py
```diff
@@ -126,19 +145,28 @@ def get_response_from(self, uri):
         Sends a GET request to the Presto server at the specified next_uri
         and updates the response
         """
-        try:
-            conn = urlopen(uri, None, URL_TIMEOUT_MS)
-            answer = conn.read()
-            conn.close()
+        parts = list(urlparse.urlsplit(uri))
```
cawallin: What is this bit doing exactly? Can you make it more explicit/less hacky looking?

prestoadmin/util/base_config.py
```diff
@@ -41,8 +79,8 @@ def prompt_user(self, conf):
                                 default=conf.get(self.key, self.default),
                                 validate=self.validate)
 
-    def collect_prompts(self, l):
-        l.append((self.prompt, self.key))
+    def get_keys(self):
```
ebd2: I don't remember why I had collect_prompts in here originally, but it was unused.

prestoadmin/util/base_config.py
```diff
@@ -103,6 +140,24 @@ class BaseConfig(object):
     def __init__(self, config_path, config_items):
         self.config_path = config_path
         self.config_items = config_items
+        self.config = {}
+
+    def __getitem__(self, key):
```
ebd2: Act like a dict so we can set env.conf for standalone configs as well as yarn_slider configs.

prestoadmin/util/base_config.py
```diff
@@ -111,10 +166,18 @@ def write_conf(self, conf):
...
 
-    def get_conf_interactive(self):
-        conf = {}
+    def get_conf_interactive(self, old_conf=None):
+        if not old_conf:
+            old_conf = {}
+        conf = dict(old_conf)
         for item in self.config_items:
-            item.prompt_user(conf)
+            if isinstance(item, SetEnvConfigItem):
```
ebd2: hack, hack, hack...

prestoadmin/yarn_slider/config.py
```diff
@@ -102,15 +102,6 @@ def __init__(self):
         super(SliderConfig, self).__init__(SLIDER_CONFIG_PATH, _SLIDER_CONFIG)
         self.config = {}
 
-    def __getitem__(self, key):
```
ebd2: Moved to BaseConfig

requirements.txt
```diff
@@ -18,4 +18,5 @@ PyYAML==3.11
 overrides==0.5
 setuptools==20.1.1
 pip==8.1.2
-retrying==1.3.3
\ No newline at end of file
+retrying==1.3.3
+pyjks==0.5.1
```
ebd2: Adds native code dependency :-(

prestoadmin/prestoclient.py
```diff
@@ -191,3 +219,91 @@ def get_rows(self, num_of_rows=NUM_ROWS):
...
+        ca_file_path = self.get_pem(env.conf)
+        https_port = self.coordinator_config.get_https_port()
+        result = HTTPSCaCertConnection(
+                self.server, https_port, None, None, ca_file_path, False, URL_TIMEOUT_MS)
+        return result
+
+    def _get_local_keystore_path(self):
+        remote_keystore_path = self.coordinator_config.get_client_keystore_path()
+        keystore_name = os.path.basename(remote_keystore_path)
+        return os.path.join(get_coordinator_directory(), keystore_name)
```
cawallin: This seems like it's assuming that PA is installed on the same host as the Presto coordinator (unless get_coordinator_directory() is the config directory)

prestoadmin/prestoclient.py
```diff
@@ -191,3 +219,91 @@ def get_rows(self, num_of_rows=NUM_ROWS):
...
+        # https://www.digicert.com/ssl-support/pem-ssl-creation.htm
+        with open(pem_path, 'w') as pem_file:
+            for der_bytes in der_bytes_list:
+                pem_file.write(self._pem_string(der_bytes, type))
+        os.close(fd)
+        return pem_path
+
+    def get_pem(self, pa_config=None):
+        local_keystore_path = self._get_local_keystore_path()
+        if not os.path.isfile(local_keystore_path):
```
cawallin: What if it changes on disk? but you have an older version

prestoadmin/prestoclient.py
```diff
@@ -191,3 +219,91 @@ def get_rows(self, num_of_rows=NUM_ROWS):
...
+            self._fetch_keystore()
+
+        keystore = jks.KeyStore.load(
+                local_keystore_path,
+                self.coordinator_config.get_client_keystore_password())
+
+        if len(keystore.private_keys.items()) == 0:
+            _, private_key = keystore.private_keys.items()[0]
```
cawallin: I don't understand this logic, if len == 0, get the 0th item?

prestoadmin/prestoclient.py
cawallin: I like the idea of moving the key error to AliasConfigItem

prestoadmin/prestoclient.py
```diff
@@ -191,3 +219,91 @@ def get_rows(self, num_of_rows=NUM_ROWS):
...
+                    [cert[1] for cert in private_key.cert_chain], 'CERTIFICATE')
+
+        return self.ca_file_path
+
+    def _get_private_key(self, keystore, pa_config=None):
+        try:
+            alias = pa_config[CERTIFICATE_ALIAS]
+            return keystore.private_keys[alias]
+        except KeyError:
+            alias = prompt_choices(keystore.private_keys.keys())
```
cawallin: Does this prompt even if you have only one item now?

prestoadmin/server.py
cawallin: ok... don't suppose you want to divide this up into different commits?

prestoadmin/util/base_config.py
cawallin: What is this for? Presumably the java 8 home but I don't follow how it's needed.

prestoadmin/standalone/config.py
```diff
@@ -163,6 +166,7 @@ def set_config_loaded(self):
 
     @overrides
     def set_env_from_conf(self, conf):
+        self.config.update(conf)
```
cawallin: what does this do?

prestoadmin/util/base_config.py
cawallin: I thought we removed the yarn_slider configs?

prestoadmin/prestoclient.py
ebd2: Good catch, thanks.

prestoadmin/util/base_config.py
ebd2: Hasn't been torched out yet.

prestoadmin/prestoclient.py
ebd2: removed

prestoadmin/prestoclient.py
ebd2: Added comment

prestoadmin/configure_cmds.py
```diff
@@ -19,27 +19,24 @@
...
 from prestoadmin.util.base_config import requires_config
+from prestoadmin.util.constants import CONFIG_PROPERTIES, LOG_PROPERTIES, \
+    JVM_CONFIG, NODE_PROPERTIES
 from prestoadmin.util.fabricapi import put_secure
 from prestoadmin.util.filesystem import ensure_parent_directories_exist
 
 __all__ = ['show']
 
-
-CONFIG_PROPERTIES = "config.properties"
```
ebd2: Moved to break a circular dependency in imports. Gotta love Python.

prestoadmin/prestoclient.py
ebd2: This is a tempfile that we lazily create and download from the cluster.

prestoadmin/server.py
ebd2: I'd rather not ;-)

tests/unit/base_unit_case.py
```diff
@@ -18,8 +18,17 @@
 
 from tests.base_test_case import BaseTestCase
 
+PRESTO_CONFIG = {
+    'http-server.http.enabled': 'true',
+    'http-server.https.enabled': 'false',
+    'http-server.http.port': '8080',
+    'http-server.https.port': '7878',
```
cawallin: doesn't match with the other defaults

tests/unit/test_prestoclient.py
```diff
@@ -22,32 +22,34 @@
...
         client = PrestoClient('any_host', "")
         self.assertRaisesRegexp(InvalidArgumentError,
                                 "Username missing",
                                 client.execute_query, "any_sql")
 
-    @patch('prestoadmin.prestoclient.lookup_port')
     @patch('prestoadmin.prestoclient.HTTPConnection')
-    def test_default_request_called(self, mock_conn, mock_port):
-        mock_port.return_value = 8080
+    def test_default_request_called(self, mock_conn, mock_presto_config):
```
cawallin: need to add `@patch` here

tests/unit/test_prestoclient.py
```diff
@@ -22,32 +22,34 @@
...
 
-    def test_no_server(self):
+    def test_no_server(self, mock_presto_config):
         client = PrestoClient("", 'any_user')
         self.assertRaisesRegexp(InvalidArgumentError,
                                 "Server IP missing",
                                 client.execute_query, "any_sql")
 
-    def test_no_user(self):
+    def test_no_user(self, mock_presto_config):
```
cawallin: need to add `@patch` here

tests/unit/test_prestoclient.py
```diff
@@ -22,32 +22,34 @@
...
 class TestPrestoClient(BaseTestCase):
-    def test_no_sql(self):
+    def test_no_sql(self, mock_presto_config):
         client = PrestoClient('any_host', 'any_user')
         self.assertRaisesRegexp(InvalidArgumentError,
                                 "SQL query missing",
                                 client.execute_query, "", )
 
-    def test_no_server(self):
+    def test_no_server(self, mock_presto_config):
```
cawallin: need to add `@patch` here

tests/unit/test_prestoclient.py
```diff
@@ -58,56 +60,46 @@ def test_default_request_called(self, mock_conn, mock_port):
                                                "any_sql", headers)
         self.assertTrue(mock_conn().getresponse.called)
 
-    @patch('prestoadmin.prestoclient.lookup_port')
     @patch('prestoadmin.prestoclient.HTTPConnection')
-    def test_connection_failed(self, mock_conn, mock_port):
-        mock_port.return_value = 8080
+    def test_connection_failed(self, mock_conn, mock_presto_config):
```
cawallin: need to add `@patch` here

tests/unit/test_prestoclient.py
```diff
@@ -58,56 +60,46 @@ def test_default_request_called(self, mock_conn, mock_port):
...
         client.execute_query("any_sql")
 
         self.assertTrue(mock_conn().close.called)
         self.assertFalse(client.execute_query("any_sql"))
 
-    @patch('prestoadmin.prestoclient.lookup_port')
     @patch('prestoadmin.prestoclient.HTTPConnection')
-    def test_http_call_failed(self, mock_conn, mock_port):
-        mock_port.return_value = 8080
+    def test_http_call_failed(self, mock_conn, mock_presto_config):
```
cawallin: ditto

tests/unit/test_prestoclient.py
```diff
@@ -58,56 +60,46 @@ def test_default_request_called(self, mock_conn, mock_port):
...
 
         mock_conn.side_effect = socket.error("Error")
         self.assertFalse(client.execute_query("any_sql"))
 
-    @patch('prestoadmin.prestoclient.lookup_port')
     @patch.object(HTTPConnection, 'request')
     @patch.object(HTTPConnection, 'getresponse')
-    def test_http_answer_valid(self, mock_response, mock_request, mock_port):
-        mock_port.return_value = 8080
+    def test_http_answer_valid(self, mock_response, mock_request, mock_presto_config):
```
cawallin: ditto

tests/unit/test_prestoclient.py
```diff
@@ -58,56 +60,46 @@ def test_default_request_called(self, mock_conn, mock_port):
...
             PropertyMock(return_value=200)
         self.assertTrue(client.execute_query('any_sql'))
 
-    @patch('prestoadmin.prestoclient.lookup_port')
     @patch.object(HTTPConnection, 'request')
     @patch.object(HTTPConnection, 'getresponse')
     def test_http_answer_not_json(self, mock_response,
-                                  mock_request, mock_port):
-        mock_port.return_value = 8080
+                                  mock_request, mock_presto_config):
```
cawallin: ditto

tests/unit/test_prestoclient.py
```diff
@@ -58,56 +60,46 @@ def test_default_request_called(self, mock_conn, mock_port):
...
             PropertyMock(return_value=200)
         self.assertRaisesRegexp(ValueError, 'No JSON object could be decoded',
                                 client.execute_query, 'any_sql')
 
-    @patch('prestoadmin.prestoclient.lookup_port')
     @patch.object(PrestoClient, 'get_response_from')
     @patch.object(PrestoClient, 'get_next_uri')
-    def test_retrieve_rows(self, mock_uri, mock_get_from_uri, mock_port):
-        mock_port.return_value = 8080
+    def test_retrieve_rows(self, mock_uri, mock_get_from_uri, mock_presto_config):
```
cawallin: ditto

tests/unit/test_prestoclient.py
```diff
@@ -137,11 +129,9 @@ def test_retrieve_rows(self, mock_uri, mock_get_from_uri, mock_port):
         self.assertEqual(client.get_rows(), expected_row)
         self.assertEqual(client.next_uri, "")
 
-    @patch('prestoadmin.prestoclient.lookup_port')
     @patch.object(PrestoClient, 'get_response_from')
     @patch.object(PrestoClient, 'get_next_uri')
-    def test_append_rows(self, mock_uri, mock_get_from_uri, mock_port):
-        mock_port.return_value = 8080
+    def test_append_rows(self, mock_uri, mock_get_from_uri, mock_presto_config):
```
cawallin: ditto

tests/unit/test_prestoclient.py
```diff
@@ -162,7 +152,7 @@ def test_append_rows(self, mock_uri, mock_get_from_uri, mock_port):
 
     @patch.object(PrestoClient, 'get_response_from')
     @patch.object(PrestoClient, 'get_next_uri')
-    def test_limit_rows(self, mock_uri, mock_get_from_uri):
+    def test_limit_rows(self, mock_uri, mock_get_from_uri, mock_presto_config):
```
cawallin: ditto

tests/unit/test_prestoclient.py
```diff
@@ -173,19 +163,9 @@ def test_limit_rows(self, mock_uri, mock_get_from_uri):
...
-        mock_urlopen.return_value = mock_resp
-        mock_resp.read.return_value = '{"message": "ok!"}'
-
-        client.get_response_from('any_uri')
-        self.assertEqual(client.response_from_server, {"message": "ok!"})
-
     @patch('prestoadmin.prestoclient.HTTPConnection')
     @patch('prestoadmin.util.remote_config_util.sudo')
-    def test_execute_query_get_port(self, sudo_mock, conn_mock):
+    def test_execute_query_get_port(self, sudo_mock, conn_mock, mock_presto_config):
```
cawallin: ditto

prestoadmin/util/httpscacertconnection.py
```diff
@@ -0,0 +1,36 @@
+import socket
+import ssl
+import httplib
+
+# Adapted from http://code.activestate.com/recipes/577548-https-httplib-client-connection-with-certificate-v/
+# BSD-licensed.
+
+
+class HTTPSCaCertConnection(httplib.HTTPSConnection):
+    """ Class to make a HTTPS connection, with support for full client-based SSL Authentication"""
```
ebd2: No, we need this because 2.6.x and < 2.7.9 don't support it. This sets up the ca-certs (using the one from the private key) so that cert validation passes on new versions of python, while *also* working on old versions.

tests/unit/base_unit_case.py
ebd2: Do we care? It's test data

prestoadmin/util/httpscacertconnection.py
```diff
@@ -0,0 +1,36 @@
...
+            This is needed to pass cert_reqs=ssl.CERT_REQUIRED as parameter to ssl.wrap_socket(),
+            which forces SSL to check server certificate against our client certificate.
+        """
+        sock = socket.create_connection((self.host, self.port), self.timeout)
+        if self._tunnel_host:
+            self.sock = sock
+            self._tunnel()
+        # If there's no CA File, don't force Server Certificate Check
+        if self.ca_file:
+            self.sock = ssl.wrap_socket(sock, self.key_file, self.cert_file, ca_certs=self.ca_file,
```
ebd2: cert_file is a certificate that the client uses to prove its own identity. We aren't using that functionality.
ca_file is a list of trusted certificates (more or less). We're passing it here to tell ssl that we trust the server even though it's (probably) using a self-signed cert.

ebd2: And no, I haven't tested with a non-self-signed cert. Let me know if you have one for a random test cluster :-b

prestoadmin/standalone/config.py
```diff
@@ -163,6 +164,7 @@ def set_config_loaded(self):
 
     @overrides
     def set_env_from_conf(self, conf):
+        self.config = conf
```
ebd2: base_config uses it in __set_item__, etc.

prestoadmin/prestoclient.py
```diff
@@ -126,19 +143,28 @@ def get_response_from(self, uri):
...
-        except (HTTPError, URLError) as e:
-            _LOGGER.error("Error opening the presto response uri: " +
-                          str(e.reason))
+        conn = self._get_connection()
+        conn.request("GET", location)
+        response = conn.getresponse()
+
+        if response.status != 200:
+            conn.close()
+            _LOGGER.error("Connection error: " +
```
ebd2: Server and path or just server?

cawallin: whatever's in the "location" variable is what I was thinking. server probably would also be helpful

prestoadmin/prestoclient.py
```diff
@@ -191,3 +217,82 @@ def get_rows(self, num_of_rows=NUM_ROWS):
...
+
+    def _get_connection(self):
+        if self.coordinator_config.use_https():
+            return self._get_https_connection()
+        else:
+            return HTTPConnection(self.server, self.port, False, URL_TIMEOUT_MS)
+
+    def _get_https_connection(self):
+        ca_file_path = self._get_pem()
+        https_port = self.coordinator_config.get_https_port()
```
ebd2: Fixed the whole mutable port mess.

prestoadmin/util/presto_config.py
```diff
@@ -0,0 +1,87 @@
...
+from fabric.operations import get
+from fabric.state import env
+from prestoadmin.config import get_conf_from_properties_data
+from prestoadmin.util.constants import REMOTE_CONF_DIR, CONFIG_PROPERTIES
+
+HTTP_ENABLED_KEY = 'http-server.http.enabled'
+HTTPS_ENABLED_KEY = 'http-server.https.enabled'
+HTTP_PORT_KEY = 'http-server.http.port'
+HTTPS_PORT_KEY = 'http-server.https.port'
+CLIENT_KEYSTORE_PATH_KEY = 'http-server.https.keystore.path'
```
cawallin: We probably should use the client keystore, and the config names changed recently:
internal-communication.https.keystore.path=<keystore path>
internal-communication.https.keystore.key=<keystore password>
(from https://github.com/Teradata/presto/blob/sprint-45/presto-docs/src/main/sphinx/security/internal-communication.rst)

ebd2: Per offline discussion, we need to make this match whatever is on the upcoming TD release.
If it changes when this lands on FB master, we'll have yet another place where we have a version dependency.

prestoadmin/util/httpscacertconnection.py
ebd2: @cawallin tested; it works

prestoadmin/prestoclient.py
```diff
@@ -191,3 +217,82 @@ def get_rows(self, num_of_rows=NUM_ROWS):
...
+    def _get_local_keystore_path(self):
+        remote_keystore_path = self.coordinator_config.get_client_keystore_path()
+        keystore_name = os.path.basename(remote_keystore_path)
+        return os.path.join(get_coordinator_directory(), keystore_name)
+
+    def _fetch_keystore_data(self):
+        if not self.keystore_data:
+            remote_keystore_path = self.coordinator_config.get_client_keystore_path()
+            keystore_data = StringIO()
+            get(remote_keystore_path, keystore_data, use_sudo=True)
```
ebd2: Useful errors now happen in both cases; presto_config calls `error()` if a key is missing, and fabric.get shows an error if the file is missing.

prestoadmin/prestoclient.py
```diff
@@ -42,15 +42,19 @@
...
-        self.port = port
         self.coordinator_config = PrestoConfig.coordinator_config()
+        self.port = PrestoClient._get_configured_port(self.coordinator_config)
+
+        # mutable stuff
         self.ca_file_path = ""
         self.keystore_data = ""
         self.rows = []
         self.next_uri = ''
+
```
cawallin: don't add whitespace here

prestoadmin/util/presto_config.py
```diff
@@ -45,28 +45,38 @@ class PrestoConfig:
...
-        return PrestoConfig(presto_config_dict)
+        return PrestoConfig(presto_config_dict, config_path, config_host)
 
     def _lookup(self, key):
-        return self.config_properties.get(key, self.default_config[key])
+        result = self.config_properties.get(key, self.default_config[key])
+        if not result:
+            raise Exception(
+                    "Key %s is not configured in coordinator configuration"
+                    "%s on host %s and has no default" %
```
cawallin: Maybe we should remove the default for CLIENT_KEYSTORE_PATH_KEY/CLIENT_KEYSTORE_PASSWORD_KEY/?

prestoadmin/util/presto_config.py
```diff
@@ -72,7 +73,7 @@ def coordinator_config():
     def _lookup(self, key):
         result = self.config_properties.get(key, self.default_config[key])
         if not result:
-            raise Exception(
+            error(
```
cawallin: are there any other places where you raise an Exception rather than doing error?

ebd2:
Rebased and merged.

PR #290, opened 2016-12-05
Edit install instructions to reflect that sudo is not required

petroav:


docs/installation/presto-admin-installation.rst
```diff
@@ -38,25 +38,24 @@ of the nodes in the cluster.
...
-with sudo when the presto-admin user is non-root.
-
-By default, the ``presto-admin`` config and log directory locations are
-configured to be at ``~/.prestoadmin`` and ``~/.prestoadmin/log``, respectively.  This
-can be changed by modifying the environment variables, PRESTO_ADMIN_CONFIG_DIR and
-PRESTO_ADMIN_LOG_DIR.
+executable ``presto-admin`` script. By default, the ``presto-admin`` config and log
+directory locations are configured to be ``~/.prestoadmin`` and ``~/.prestoadmin/log``,
+respectively.  This can be changed by modifying the environment variables,
+PRESTO_ADMIN_CONFIG_DIR and PRESTO_ADMIN_LOG_DIR. The installation script will also create
```
rschlussel: I would say  "The installation script will create the appropriate configuration and log directories for presto-admin".  I'd leave out the second part of the sentence about using sudo.  I think it's clear enough that if you're doing something that requires sudo privileges, you should use sudo.  You'll get an error message if you mess up.  and i think the docs should be optimized for being clear for the default configurations, for which you definitely don't want to use sudo or those directories will be created in the wrong place.

petroav: OK, I agree, don't have a strong opinion about this. Fixed in https://github.com/prestodb/presto-admin/pull/290/commits/d3c854935188b202791a0605c37f98b10565437a

petroav:
@rschlussel do you have any other comments?

docs/installation/presto-admin-installation.rst
```diff
@@ -38,25 +38,22 @@ of the nodes in the cluster.
...
+the directories pointed to by PRESTO_ADMIN_CONFIG_DIR and PRESTO_ADMIN_LOG_DIR.
```
petroav: Looks like all directories are created with the `-p` flag so we should be good.

```
CONF_DIR=${PRESTO_ADMIN_CONFIG_DIR:-~/.prestoadmin}
mkdir -p $CONF_DIR

LOG_DIR=${PRESTO_ADMIN_LOG_DIR:-$CONF_DIR/log}
mkdir -p $LOG_DIR

CATALOG_DIR=$CONF_DIR/catalog
mkdir -p $CATALOG_DIR

COORDINATOR_DIR=$CONF_DIR/coordinator
mkdir -p $COORDINATOR_DIR

WORKERS_DIR=$CONF_DIR/workers
mkdir -p $WORKERS_DIR
```

docs/installation/presto-admin-installation.rst
```diff
@@ -38,25 +38,22 @@ of the nodes in the cluster.
...
 
 3. Verify that ``presto-admin`` was installed properly by running the
```
petroav: How about: "by running the following command:" and then having a code block with "./presto-admin --help"

ebd2: Sounds fine to me.

PR #292, opened 2016-12-07
Refactor stop_host method to use topology

petroav:
- Stopping containers within a test enabled race
  conditions. This change refactors the stop_host
  method to substitute the presto-admin topology
  with a fake hostname effectively shutting down
  one of the containters.


tests/product/test_configuration.py
```diff
@@ -195,31 +195,31 @@ def test_configuration_deploy_using_dash_x_coord_worker(self):
...
+        bad_host = self.cluster.get_down_hostname()
         good_hosts = [self.cluster.internal_master,
                       self.cluster.internal_slaves[1],
                       self.cluster.internal_slaves[2]]
-        topology = {'coordinator': internal_bad_host,
+        topology = {'coordinator': bad_host,
                     'workers': good_hosts}
         self.upload_topology(topology)
-        self.cluster.stop_host(bad_host)
+        self.cluster.stop_host(self.cluster.slaves[0])
```
rschlussel: you've already put the bad host name in the topology, so stopping the host should really do anything. Maybe we don't need stop_host at all, just upload bad worker/coordinator topologies all around.   In any case it should be one or the other.

tests/base_cluster.py
```diff
@@ -84,20 +85,39 @@ def get_ip_address_dict(self):
...
+        """Stop a host by editing the presto-admin topology to point
+        to a non-existent host as provided by get_down_hostname() and
+        get_down_ip()"""
+        if host_name not in self.all_hosts():
+            raise Exception('Must specify external hostname to stop_host')
+
+        # Change the topology to something that doesn't exist
+        ips = self.get_ip_address_dict()
+        down_hostname = self.get_down_hostname()
+        self.exec_cmd_on_host(
```
rschlussel: Why do we need all three of these sed commands? Doesn't the topology usually just contain the internal host name?

petroav:
@rschlussel updated version that addresses your comments. I also added a commit that fixed the `Connection reset` errors I was seeing while building the offline installer.

rschlussel:
travis failures are real.  It looks like you missed a few usages of stop_host

rschlussel:
looks good % fixing the test failures.

rschlussel:
Can you also revert the timeout increase (and make sure that this fixes the issue)?

rschlussel:
@petroav what's the status of this?

petroav:
@rschlussel I put this on pause since we were seeing all those `UnixPoolConnectionError` I thought I would wait until we fixed the intermittent errors before going forward with this. I just saw the Travis has actually been very stable for presto-admin so I might rebase this.

PR #293, opened 2016-12-09
Add release notes for release 2.0

rschlussel:


docs/release/release-2.0.rst
```diff
@@ -0,0 +1,40 @@
...
+
+New Features
+------------
+* Make presto-admin log and configuration directories configurable. They can be
+  set using the environment variables ``PRESTO_ADMIN_LOG_DIR`` and
+  ``PRESTO_ADMIN_CONFIG_DIR``.
+* Change the default configuration directory to ``~/.prestoadmin`` and the
+  default log directory to ``~/.prestoadmin/log``
+* Remove requirement for running and installing presto-admin with sudo.
+  Presto-Admin still uses sudo to execute commands on the Presto nodes.
```
petroav: I'm confused by this line.

docs/release/release-2.0.rst
```diff
@@ -0,0 +1,40 @@
+===========
+Release 2.0
+===========
+
+New Features
+------------
+* Make presto-admin log and configuration directories configurable. They can be
+  set using the environment variables ``PRESTO_ADMIN_LOG_DIR`` and
+  ``PRESTO_ADMIN_CONFIG_DIR``.
```
petroav: Some bullet points end in periods, some don't. I would prefer every bullet point to have a period at the end but I don't feel strongly as long as we're consistent.

docs/release/release-2.0.rst
```diff
@@ -0,0 +1,40 @@
...
+* Allow specifying which python to use as an argument to the presto-admin
+  installation script
+* Add ``G1HeapRegionSize=32M`` to the jvm.config defaults as suggested by the
+  Presto documentation
+
+Bug Fixes
+---------
+* Keep the ``node.id`` in Presto's ``node.properties`` file consistent across
+  configuration updates
+* Give Presto configuration files ``644`` permissions
```
cawallin: As I mentioned offline, I don't think we should have this change

docs/release/release-2.0.rst
```diff
@@ -0,0 +1,40 @@
...
+  default log directory to ``~/.prestoadmin/log``
+* Remove requirement for running and installing presto-admin with sudo.
+  Presto-Admin still uses sudo to execute commands on the Presto nodes.
+* Rename the ``connectors`` directory to ``catalog`` to match the Presto
+  nomenclature
+* Rename the ``connector add`` and ``connector remove``  commands to
+  ``catalog add`` and ``catalog remove``
+* Add experimental support for connecting to a Presto server with internal
+  communication via HTTPS and LDAP, where the HTTP connection is disabled
+* Allow specifying which python to use as an argument to the presto-admin
```
petroav: -> "which python interpreter"

docs/release/release-2.0.rst
petroav: -> Change Presto configuration files to have 644 permissions.

Give sounds weird.

docs/release/release-2.0.rst
rschlussel: I was trying to say that presto-admin still uses sudo internally.  I can just leave that bit out if it's more confusing than helpful

docs/release/release-2.0.rst
rschlussel: getting rid of this anyway

docs/release/release-2.0.rst
petroav: How does it use sudo internally? I'm still confused :/

docs/release/release-2.0.rst
rschlussel: but i changed the line below

docs/release/release-2.0.rst
rschlussel: done

rschlussel:
addressed comments

docs/release/release-2.0.rst
```diff
@@ -0,0 +1,38 @@
...
+===========
+
+New Features
+------------
+* Make presto-admin log and configuration directories configurable. They can be
+  set using the environment variables ``PRESTO_ADMIN_LOG_DIR`` and
+  ``PRESTO_ADMIN_CONFIG_DIR``.
+* Change the default configuration directory to ``~/.prestoadmin`` and the
+  default log directory to ``~/.prestoadmin/log``.
+* Remove the requirement for running and installing presto-admin with sudo.
```
cawallin: I disagree with Anton, I think you should mention that you need sudo access on all the nodes. A lot of people (including Apple) are confused about this.

docs/release/release-2.0.rst
```diff
@@ -10,6 +10,7 @@ New Features
 * Change the default configuration directory to ``~/.prestoadmin`` and the
   default log directory to ``~/.prestoadmin/log``.
 * Remove the requirement for running and installing presto-admin with sudo.
```
cawallin: suggest:
"Remove the requirement for running and installing the `presto-admin` program with sudo. Sudo access is still required for all Presto nodes, for example to install RPMs.

rschlussel: didn't include the bit about to install RPM's because we use sudo for just about every command we run for Presto

cawallin: I think we need some sort of explanation though, b/c it's confusing that we say that we "removed the requirement for sudo" when you still do need sudo.

rschlussel: How about "The user specified in config.json still needs sudo access on the Presto nodes in order to execute commands like installing the rpm and setting permissions on the configuration files."

cawallin: +1

cawallin: RPM capitalized

rschlussel:
merged

PR #296, opened 2016-12-12
Fix the markdown syntax error. 

yuokada:
Hi, 
Fix  code-highlights.

----


ebd2:
Thanks for cleaning up the documentation!

PR #298, opened 2016-12-13
Add unique temp dir when issuing get in parallel

cawallin:
Fabric's get command, when using sudo, copies the file
to a location that is the sha1 of the filename, then ftps
the file. This doesn't work in parallel because the filename
is not unique so it gets deleted halfway through some of
the operations.

### DIFF
--- a/prestoadmin/util/presto_config.py
+++ b/prestoadmin/util/presto_config.py
@@ -16,7 +16,7 @@ import os
 from StringIO import StringIO
 
 from fabric.context_managers import settings, hide
-from fabric.operations import get, sudo
+from fabric.operations import get, run
 from fabric.state import env
 from fabric.utils import error
 from prestoadmin.config import get_conf_from_properties_data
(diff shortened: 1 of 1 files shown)


prestoadmin/util/presto_config.py
```diff
@@ -66,12 +66,17 @@ def coordinator_config():
         try:
             data = StringIO()
             with settings(host_string='%s@%s' % (env.user, config_host)):
-                get(config_path, data, use_sudo=True)
+                with hide('stderr', 'stdout'):
+                    temp_dir = sudo('mktemp -d /tmp/prestoadmin.XXXXXXXXXXXXXX')
```
rschlussel: does this need sudo?  At some point we're going to need to work on purging unnecessary sudo calls.  If it does need sudo, maybe just generate a random string here, and pass it in as temp_dir

cawallin: good call, it doesn't need sudo

rschlussel: unless it needs to already exist, I guess.

cawallin: unfortunately it already needs to exist, fabric doesn't create the directory

PR #300, opened 2016-12-21
use_ldap should return true only when HTTP is disabled

ebd2:
@cawallin @anusudarsan

### DIFF
--- a/prestoadmin/util/presto_config.py
+++ b/prestoadmin/util/presto_config.py
@@ -19,6 +19,7 @@ from fabric.context_managers import settings, hide
 from fabric.operations import get, run
 from fabric.state import env
 from fabric.utils import error
+
 from prestoadmin.config import get_conf_from_properties_data
 from prestoadmin.util.constants import REMOTE_CONF_DIR, CONFIG_PROPERTIES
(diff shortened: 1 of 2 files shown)


prestoadmin/util/presto_config.py
```diff
@@ -60,25 +60,24 @@ def __init__(self, config_properties, config_host, config_path):
...
         try:
             data = StringIO()
             with settings(host_string='%s@%s' % (env.user, config_host)):
-                with hide('stderr', 'stdout'):
-                    temp_dir = run('mktemp -d /tmp/prestoadmin.XXXXXXXXXXXXXX')
-                try:
-                    get(config_path, data, use_sudo=True, temp_dir=temp_dir)
-                finally:
-                    run('rm -r %s' % temp_dir)
+                get(config_path, data)
```
cawallin: As I noted offline, you're probably not based on the most recent code.

ebd2: I am not. Fixing it now.

ebd2: Nothing to see here :-D

PR #305, opened 2017-01-12
Make version matching less restrictive

rschlussel:
This is needed to allow versions like 0.162.SNAPSHOT.t.SNAPSHOT

Now if words appear between any of the numbered bits, we just ignore them. 

I also changed the responsibility for splitting the `-` to be part of when we split the version on `.`.

### DIFF
--- a/prestoadmin/util/version_util.py
+++ b/prestoadmin/util/version_util.py
@@ -19,102 +19,55 @@ Stuff to handle version ranges.
 import re
 
 TD_VERSION = re.compile(r'^\d+t$')
-ANCIENT_TAGGED_VERSION = re.compile(r'^\d+t?-[A-Z]+$')
 
 
 def split_version(version_string):
-    return version_string.strip().split('.')
+    # We split on '.' and '-' because ancient tagged versions had the tag
+    # delimited by a '-'
+    return re.split('\.|-', version_string.strip())
 
 
-# New versions of the Teradata RPM are of the form 0.148.t.x.y, so the 't'
-# can be not attached to the preceding version number, and we want to
-# preserve it.
-def intOrT(x):
+def get_int_or_t(x):
     try:
         return int(x)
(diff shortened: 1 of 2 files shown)


prestoadmin/util/version_util.py
```diff
@@ -19,102 +19,58 @@
...
 
 
 def split_version(version_string):
-    return version_string.strip().split('.')
+    # We split on '.' and '-' because ancient tagged versions had the tag
+    # delimited by a '-'
+    return re.split('\.|-', version_string.strip())
 
 
 # New versions of the Teradata RPM are of the form 0.148.t.x.y, so the 't'
```
ebd2: Comment is out of date for version 0.123t

tests/unit/util/test_version_util.py
```diff
@@ -112,31 +112,20 @@ def test_trailing_non_numeric(self):
...
 
     def test_teradata_version(self):
         self.assertEqual(
             (0, '115t'), strip_tag(('0', '115t')))
         self.assertEqual(
             (0, '123t'), strip_tag(('0', '123t', 'SNAPSHOT')))
 
-    def test_non_trailing_non_numeric(self):
-        self.assertRaises(
-            ValueError, strip_tag, ('1', 'TWO', '3'))
```
ebd2: Test for new behavior specified by comment here: https://github.com/prestodb/presto-admin/pull/305/files#diff-c6d431ed89d19e8713c881bbf37a396aR58

prestoadmin/util/version_util.py
rschlussel: I think the comment is talking about post 0.123t, but it's still not that helpful.  I'll get rid of it.

rschlussel:
@ebd2 addressed comments

rschlussel:
merged

PR #306, opened 2017-01-13
Remove rpm_cache_dir on tear_down

kokosing:
Remove rpm_cache_dir on tear_down

Method presto_installer#copy_presto_rpm_to_master first check if there
is no rpm already on master then it is copying the rpm file.

This changes makes it possible that the same configurable cluster can be
used for another tests with different rpm file.


cawallin:
I think this will force you to upload the RPM for every test, which will make the tests (which are already slow) much slower

cawallin:
A better choice for running tests with multiple RPMs would be to have a separate RPM cache, as specified in the config file

PR #307, opened 2017-01-13
Upload rpm when remote rpm is older than local

kokosing:
Upload rpm when remote rpm is older than local

Method presto_installer#copy_presto_rpm_to_master first check if there
is no rpm already on master then it is copying the rpm file.

This changes makes it possible that the same configurable cluster can be
used for another tests with different rpm file.


rschlussel:
This will mean that when you that a new rpm will be uploaded for every test that you run.  This will make the tests take unreasonably long (more than they are already)

PR #309, opened 2017-01-27
Make product test to be runnable on Teradata supplied hardware

kokosing:

### DIFF
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -623,10 +623,12 @@ def check_server_status():
 @retry(stop_max_delay=RETRY_TIMEOUT * 1000, wait_fixed=5000, retry_on_result=lambda result: result is False)
 def query_server_for_status(client, node_id):
     try:
-        client.execute_query(SYSTEM_RUNTIME_NODES)
+        rows = client.run_sql(SYSTEM_RUNTIME_NODES)
+        if rows is not None:
+            return _is_in_rows(node_id, rows)
     except ConfigurationError as e:
         _LOGGER.warn(e)
-    return _is_in_rows(node_id, client.get_rows())
+    return False
 
 
 def _is_in_rows(value, rows):
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -782,6 +782,8 @@ def collect_node_information():
 
 
 def get_status_from_coordinator():
+    if check_presto_version() != '':
+        return
     with closing(PrestoClient(get_coordinator_role()[0], env.user)) as client:
         try:
             coordinator_status = run_sql(client, SYSTEM_RUNTIME_NODES)
--- a/prestoadmin/server.py
+++ b/prestoadmin/server.py
@@ -636,17 +636,6 @@ def _is_in_rows(value, rows):
     return False
 
 
-def run_sql(client, sql):
-    status = client.execute_query(sql)
-    if status:
-        return client.get_rows()
-    else:
-        # TODO: Check if we can get some error cause from server response and
-        # log that to the user
-        _LOGGER.error('Querying server failed')
-        return []
-
-
 def execute_catalog_info_sql(client):
     """
     Returns [[catalog_name], [catalog_2]..] from catalogs system table
@@ -654,7 +643,7 @@ def execute_catalog_info_sql(client):
     Parameters:
         client - client that executes the query
     """
-    return run_sql(client, CATALOG_INFO_SQL)
+    return client.run_sql(CATALOG_INFO_SQL)
 
 
 def execute_external_ip_sql(client, uuid):
@@ -666,7 +655,7 @@ def execute_external_ip_sql(client, uuid):
         client - client that executes the query
         uuid - node_id of the node
     """
-    return run_sql(client, EXTERNAL_IP_SQL % uuid)
+    return client.run_sql(EXTERNAL_IP_SQL % uuid)
 
 
 def get_sysnode_info_from(node_info_row, state_transform):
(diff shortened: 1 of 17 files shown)


tests/product/test_status.py
```diff
@@ -111,7 +111,7 @@ def test_connection_to_worker_lost(self):
             topology, self.cluster.internal_slaves[1])
         self.check_status(status_output, statuses)
 
-    def test_status_port_not_8080(self):
+    def test_status_port_not_28384(self):
```
rschlussel: I think this test is no longer necessary.  The point was to make sure 8080 wasn't hardcoded (since at some point there was a bug where it had been somewhere).

prestoadmin/util/presto_config.py
```diff
@@ -66,12 +66,11 @@ def from_file(config, config_path=None, config_host=None):
         return PrestoConfig(presto_config_dict, config_path, config_host)
 
     @staticmethod
-    def coordinator_config():
+    def config(host, user):
```
rschlussel: 1. Can you make the commit message more clear ("give inject host and user" + the description).  I know what this does, and still don't understand what the commit message is saying :) 

2.  This change doesn't seem to be used at all.  You have a different solution in the commit "Do not call fabric based function from product test".   I like this approach better because you can't accidentally call the wrong version.

tests/product/test_collect.py
```diff
@@ -160,7 +163,10 @@ def get_query_id(self, sql, host=None):
         ips = self.cluster.get_ip_address_dict()
         if host is None:
             host = self.cluster.master
-        client = PrestoClient(ips[host], 'root')
+        config_path = os.path.join(REMOTE_CONF_DIR, CONFIG_PROPERTIES)
+        config = self.cluster.exec_cmd_on_host(host, 'cat ' + config_path)
+        user = 'root'
+        client = PrestoClient(ips[host], user, PrestoConfig.from_file(StringIO(config), config_path, host))
```
rschlussel: As noted, this seems to be an alternative solution to the one that passed the user and host to the client's config-getting function.  I prefer the other way, but definitely we only need one solution.

prestoadmin/util/presto_config.py
```diff
@@ -52,7 +51,7 @@ class PrestoConfig:
         LDAP_CLIENT_PASSWORD_KEY: None
     }
 
-    def __init__(self, config_properties, config_host, config_path):
+    def __init__(self, config_properties, config_path, config_host):
```
rschlussel: Can you change the commit message to "change order of arguments for PrestoConfig"

prestoadmin/util/presto_config.py
```diff
@@ -68,21 +68,17 @@ def from_file(config, config_path=None, config_host=None):
...
-                    get(config_path, data, use_sudo=True, temp_dir=temp_dir)
-                finally:
-                    run('rm -r %s' % temp_dir)
-
-            data.seek(0)
-            return PrestoConfig.from_file(data, config_path, host)
-        except Exception as e:
-            _LOGGER.info('Could not find Presto config.')
-            return PrestoConfig(None, config_path, host)
+        data = StringIO()
```
rschlussel: 1. instead of propagating  exceptions for things that could be user facing, you should use the error() function to give a nicer error message.  
2.  This was here because the client gets used for `server status`, and if a user calls `server status` without Presto installed, we don't want an exception.  We want to report back to the user that Presto is not installed. 
However, I think this change is a good one.  Status should be checking whether Presto is installed first, and not call out to the client if it knows that it's not (now the check happens somewhere in the middle of everything). The code for status is a bit messy, but for your change to work correctly, you would need to change `server.get_status_from_coordinator()` to check that Presto is installed before trying to query anything else.

tests/product/test_collect.py
kokosing: this solution works in product tests. Passing `host` and `user` is not enough for product tests as product test is not able to call things like (part of `coordinator_config()`):

```
                with hide('stderr', 'stdout'):
                    temp_dir = run('mktemp -d /tmp/prestoadmin.XXXXXXXXXXXXXX')
```
fabric methods like `run` requires fabric to properly initialized, so it cannot be simply called from product tests

prestoadmin/util/presto_config.py
kokosing: As I understand I had two options. So I chose 2. Please see https://github.com/prestodb/presto-admin/pull/309/commits/19dce92cc4d205d61956214e0328af11d8242f2b, there I check if presto is installed before checking its status.

prestoadmin/util/presto_config.py
kokosing: I chose number 2. I removed this commit.

prestoadmin/server.py
```diff
@@ -782,6 +782,8 @@ def collect_node_information():
 
 
 def get_status_from_coordinator():
+    if check_presto_version() != '':
```
rschlussel: This isn't quite right (this checks that presto is installed on your presto-admin node, and if not will print nothing out for the `server status` command). 

 But looking at this function again, I think it may be ok as it was.  Maybe see if the test_status tests fail from having the client raise an exception when it can't generate a config.  If it does I can give you more direction on how to rewrite it.

prestoadmin/util/presto_config.py
```diff
@@ -69,21 +69,17 @@ def from_file(config, config_path=None, config_host=None):
...
+        with settings(host_string='%s@%s' % (env.user, config_host)):
+            with hide('stderr', 'stdout'):
+                temp_dir = run('mktemp -d /tmp/prestoadmin.XXXXXXXXXXXXXX')
+            try:
+                get(config_path, data, use_sudo=True, temp_dir=temp_dir)
+            finally:
+                run('rm -r %s' % temp_dir)
+
+        data.seek(0)
+        return PrestoConfig.from_file(data, config_path, config_host)
```
rschlussel: What I meant is I think we should catch the exception here and call the error function with a nice message.

```
except Exception as e:
    error('Could not get coordinator config', exception=e)
```

kokosing:
@rschlussel I am going to close this PR. I will now open one PR per each change. presto-admin is very fragile and making something to work on travis as well on jenkins is difficult. Hope with that approach something will get in.

The first PR is here: https://github.com/prestodb/presto-admin/pull/314

PR #310, opened 2017-02-06
Execute “rpm  -U" when "server upgrade" task

yuokada:
I don't want to use `--nodeps` option when upgrade a rpm package  because it's dangerous.

### DIFF
--- a/prestoadmin/package.py
+++ b/prestoadmin/package.py
@@ -125,10 +125,12 @@ def rpm_upgrade(rpm_name):
     if not package_name.succeeded:
         abort("Corrupted RPM file: %s" % rpm_path)
 
-    if _rpm_uninstall(package_name).succeeded:
-        if _rpm_install(rpm_path).succeeded:
-            print("Package upgraded successfully on: " + env.host)
+    if _rpm_upgrade(rpm_path).succeeded:
+        print("Package upgraded successfully on: " + env.host)
 
+
+def _rpm_upgrade(package_name):
+    return sudo('rpm -U %s%s' % (_nodeps_rpm_option(), package_name))
 
 @task
 @runs_once
--- a/tests/unit/test_package.py
+++ b/tests/unit/test_package.py
@@ -51,10 +51,9 @@ class TestPackage(BaseUnitCase):
         mock_sudo.assert_called_with('rpm -i --nodeps '
                                      '/opt/prestoadmin/packages/test.rpm')
 
-    @patch('prestoadmin.package._rpm_uninstall')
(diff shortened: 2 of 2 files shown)


cawallin:
Thanks for the PR! I think the product test failures come because our dummy rpm (the stripped down, small RPM we use for upgrade that is checked into the repo) was created against Presto 0.153, and so it can't upgrade to the most recent version. We unfortunately don't have the process of making a dummy rpm documented, so I can make one for this and give it some extremely high version so we don't run into this again (like Presto 10.100 or something)

yuokada:
@cawallin Hi, Thanks for reply. 
I have understantd that you have only one version rpm for testing. 
And, "rpm -U ~" failed because installed rpm and Upgrade rpm is same one.
Is this correct?
So, Should I wait for your update, Shouldn't I ?

cawallin:
@yuokada we're working on getting a new dummy RPM for these tests. The problem is, before the tests uninstalled the old RPM and then installed a dummy RPM, but now they use rpm -U. When you use rpm -U, you need to have the version of the new RPM higher than the version of the old RPM, but the dummy RPM's version was 0.153. We're working on getting you a new dummy RPM now.

cawallin:
@yuokada -- we have a dummy rpm that works, I sent it to you on slack.
With this RPM, the only test failure is with test_rolling_upgrade, which is redundant with your -U change. You can remove it.

cawallin:
Also, please remove the documentation about --nodeps from the upgrade() function in server.py, since it's not allowed in server upgrade now. (it's still allowed for package install, which seems fine by me)

amrutagokhale:
@yuokada We have uploaded the new dummy rpm to the master branch. If you rebase your PR against the latest master, the failures should go away.

yuokada:
> "[slave2] out: 	package presto-server-rpm-0:100.100.SNAPSHOT-1.x86_64 is already installed"

HM?
https://travis-ci.org/prestodb/presto-admin/jobs/202098819
```
[slave2] out: Stopping presto 
[slave2] out: Not running
[slave2] out: 
[slave2] out: /tmp/presto_config-FdEuLFE.tar
[slave2] out: 
Deploying rpm on slave2...
Package deployed successfully on: slave2
[slave2] out: 	package presto-server-rpm-0:100.100.SNAPSHOT-1.x86_64 is already installed
[slave2] out: 
Fatal error: [slave1] sudo() received nonzero return code 1 while executing!
Requested
```

yuokada:
> We'll merge it shortly.

OK,  I remove "[WIP]" in title.

amrutagokhale:
Merged, thanks!

PR #311, opened 2017-02-13
Fix some quoting in bash script

findepi:


petroav:
I think the two test failures are legit. Also I think you can probably squash your commits.

findepi:
Indeed, the problem was legit. In ```CONF_DIR="${PRESTO_ADMIN_CONFIG_DIR:-~/.prestoadmin}"```, if `$PRESTO_ADMIN_CONFIG_DIR` is not set, `CONF_DIR` is set to `~/.prestoadmin`, i.e. no expansion occurs. I checked that the assignment without apostrophes works even when `$PRESTO_ADMIN_CONFIG_DIR` contains spaces, so unquoting it should be OK. I retained proper quoting in calls to `mkdir` though.

findepi:
:+1:

PR #316, opened 2017-03-02
Do not use 8080 port for product tests

kokosing:
Do not use 8080 port for product tests

8080 is a commonly used port. When running product tests on real
hardware (not docker) then it is high chance that this port is already
taken.


kokosing:
This PR is extracted from https://github.com/prestodb/presto-admin/pull/309

PR #317, opened 2017-03-02
Change the order of arguments for PrestoConfig constructor

kokosing:
Change the order of arguments for PrestoConfig constructor

That way prestoadmin.util.presto_config.PrestoConfig#from_file and
prestoadmin.util.presto_config.PrestoConfig#coordinator_config create
proper PrestoConfig.


kokosing:
This PR is extracted from #309

PR #318, opened 2017-03-02
Do not check IP when validating the status

kokosing:
Do not check IP when validating the status

When product tests are run on real hardware it might be the case that
node has several interfaces. Determining which one will be actually used
is not trivial and will make this to be very fragile.


kokosing:
This PR is extracted from #309

PR #319, opened 2017-03-02
Do not supress exception

kokosing:
Do not supress exception

Returning "correct" object while in case of failure is not a good idea.
This commit fixes that.

### DIFF
--- a/prestoadmin/util/presto_config.py
+++ b/prestoadmin/util/presto_config.py
@@ -69,21 +69,17 @@ class PrestoConfig:
     def coordinator_config():
         config_path = os.path.join(REMOTE_CONF_DIR, CONFIG_PROPERTIES)
         config_host = env.roledefs['coordinator'][0]
-        try:
-            data = StringIO()
-            with settings(host_string='%s@%s' % (env.user, config_host)):
-                with hide('stderr', 'stdout'):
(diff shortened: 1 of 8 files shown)


kokosing:
Extracted from #309

prestoadmin/util/presto_config.py
```diff
@@ -69,21 +69,17 @@ def from_file(config, config_path=None, config_host=None):
...
-                    get(config_path, data, use_sudo=True, temp_dir=temp_dir)
-                finally:
-                    run('rm -r %s' % temp_dir)
-
-            data.seek(0)
-            return PrestoConfig.from_file(data, config_path, config_host)
-        except:
-            _LOGGER.info('Could not find Presto config.')
-            return PrestoConfig(None, config_path, config_host)
+        data = StringIO()
```
cawallin: If you're doing server status when Presto isn't yet installed, you won't have a presto config there yet.

rschlussel: When I looked more closely at the status code, it seemed like maybe it wasn't relevant, but product tests are failing, so presumably it is.  That means that get_status_from_coordinator() needs to be changed to isolate the calls to presto_config to only be used when you know presto is installed.

PR #322, opened 2017-03-03
Refactor prestoclient and use it in test_catalog

kokosing:
Use prestoclient in test_catalog

### DIFF
--- a/tests/product/test_collect.py
+++ b/tests/product/test_collect.py
@@ -160,13 +160,7 @@ class TestCollect(BaseProductTestCase):
         self.assertEqual(actual, expected)
 
     def get_query_id(self, sql, host=None):
-        ips = self.cluster.get_ip_address_dict()
-        if host is None:
-            host = self.cluster.master
-        config_path = os.path.join(REMOTE_CONF_DIR, CONFIG_PROPERTIES)
-        config = self.cluster.exec_cmd_on_host(host, 'cat ' + config_path)
-        user = 'root'
-        client = PrestoClient(ips[host], user, PrestoConfig.from_file(StringIO(config), config_path, host))
+        client = self.create_presto_client(self, host)
         run_sql(client, sql)
         query_runtime_info = run_sql(client, 'SELECT query_id FROM '
                                              'system.runtime.queries '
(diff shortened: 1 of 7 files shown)


tests/product/test_collect.py
```diff
@@ -160,13 +160,7 @@ def test_query_info_pa_separate_node(self):
...
 
     def get_query_id(self, sql, host=None):
-        ips = self.cluster.get_ip_address_dict()
-        if host is None:
-            host = self.cluster.master
-        config_path = os.path.join(REMOTE_CONF_DIR, CONFIG_PROPERTIES)
-        config = self.cluster.exec_cmd_on_host(host, 'cat ' + config_path)
-        user = 'root'
-        client = PrestoClient(ips[host], user, PrestoConfig.from_file(StringIO(config), config_path, host))
+        client = self.create_presto_client(self, host)
```
cawallin: There're probably unused imports you should remove.

tests/unit/test_prestoclient.py
```diff
@@ -94,118 +94,52 @@ def test_http_answer_not_json(self, mock_response,
         type(mock_response.return_value).status =\
             PropertyMock(return_value=200)
         self.assertRaisesRegexp(ValueError, 'No JSON object could be decoded',
-                                client.execute_query, 'any_sql')
-
-    @patch.object(PrestoClient, 'get_response_from')
-    @patch.object(PrestoClient, 'get_next_uri')
-    def test_retrieve_rows(self, mock_uri, mock_get_from_uri, mock_presto_config):
```
cawallin: This means we prob don't have any unit test coverage of retrieve_rows, which is significant because we'll only catch issues (even syntax issues) at runtime, but we have pretty thorough product testing so this is probably ok.

kokosing: Generally I would not much believe in mock testing of PrestoClient. I think the best would be to test `PrestoClient` against the real Presto server. There are many quirks that are very difficult to test by mock, which I think would be much easier if we could use real server. Presto Server API may change and here we would not notice this until product tests anyway.

cawallin: Yeah -- the unit testing is more about Python syntax than about catching API changes. But that's fine.

PR #326, opened 2017-03-06
Move PYTHONPATH setup from .travis.yml to Makefile

kokosing:
Move PYTHONPATH setup from .travis.yml to Makefile

Thanks to that Make commands (like test-images) is working not only on
travis, but also within the developer sandbox as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ BASE_IMAGES_TAG := $(shell awk '/base_images_tag/ \
 	{split($$NF, a, "\""); print a[2]}' base-images-tag.json)
 
 test-images: docker-images presto-server-rpm.rpm
-	python tests/product/image_builder.py $(IMAGE_NAMES)
+	PYTHONPATH=${PYTHONPATH}:$(shell pwd) python tests/product/image_builder.py $(IMAGE_NAMES)
 
 DOCKER_IMAGES := \
 	teradatalabs/centos6-ssh-oj8:$(BASE_IMAGES_TAG) \
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ services:
 env:
   global:
     - PA_TEST_ONLINE_INSTALLER=true
-    - PYTHONPATH=$PYTHONPATH:$(pwd)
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -135,7 +135,7 @@ BASE_IMAGES_TAG := $(shell awk '/base_images_tag/ \
 	{split($$NF, a, "\""); print a[2]}' base-images-tag.json)
 
 test-images: docker-images presto-server-rpm.rpm
-	python tests/product/image_builder.py $(IMAGE_NAMES)
+	PYTHONPATH=${PYTHONPATH}:$(shell pwd) python tests/product/image_builder.py $(IMAGE_NAMES)
```
cawallin: I got an error when I applied this patch locally:
ImportError: No module named docker
Makefile:138: recipe for target 'test-images' failed
make: *** [test-images] Error 1

It works both with and without the patch if I'm in my prestoadmin virtualenv

kokosing: if you have a preastoadmin installed in your virtualenv then this change is not needed.

The error you gave points that you don't have a docker installed in your virtualenv. Does this command works without this change? If so then it is strange, or I still need to learn something about python and PYTHONPATH.

kokosing: It is passing an automation so I thought it is safe

rschlussel: I also get an error with this patch, and have no problems without it.  

```
PYTHONPATH=:/home/rschlussel/presto-admin python tests/product/image_builder.py "all"
Traceback (most recent call last):
  File "tests/product/image_builder.py", line 1, in <module>
    from tests.docker_cluster import DockerCluster
ImportError: No module named docker_cluster
make: *** [test-images] Error 1
```

kokosing: I think it works for you because you have presto admin installed in your virtualenv. Can you please do `pip uninstall prestoadmin` and try again?

I think having presto-admin installed in virtualenv could be misleading as code changes wont be seen until you update presto in virtualenv (you need to do `pip install` or something similar to make build script be affected by the changes).

cawallin: Ahh, I see. If you do "python setup.py develop", you have access to the presto-admin files and pick up local changes to the code.

kokosing: thank you! that is working

PR #332, opened 2017-03-09
Use presto-admin coordinator config.properties

kokosing:
Use presto-admin coordinator config.properties

Reading actual presto coordinator config.properties may not always
succeed as it could be written as different user as user who runs
product tests and so it could not have access to read:
/etc/presto/config.properties file.

This is needed to fix presto-admin product tests on EMR.

### DIFF
--- a/tests/product/base_product_case.py
+++ b/tests/product/base_product_case.py
@@ -20,14 +20,14 @@ other utilities
 import json
 import os
 import re
-
 from StringIO import StringIO
+
 from nose.tools import nottest
 from retrying import Retrying
 
 from prestoadmin.prestoclient import PrestoClient
 from prestoadmin.util import constants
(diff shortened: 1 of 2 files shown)


tests/product/base_product_case.py
```diff
@@ -496,11 +496,11 @@ def status_node_connection_error(self, host):
 
     def create_presto_client(self, host=None):
         ips = self.cluster.get_ip_address_dict()
+        config_path = os.path.join('~', LOCAL_CONF_DIR, COORDINATOR_DIR_NAME, CONFIG_PROPERTIES)
+        config = self.cluster.exec_cmd_on_host(self.cluster.master, 'cat ' + config_path)
```
cawallin: You could also use sudo (use_sudo=true or something like that) to retrieve the file (that's what presto-admin does internally) but this is also fine.

PR #334, opened 2017-03-13
Use -XX:+ExitOnOutOfMemoryError instead of '-XX:OnOutOfMemoryError=kill -9 %p'

findepi:
The flag is available since JDK 8u92. Presto's docs & examples use the new flag since prestodb/presto#7056


rschlussel:
Do we care that it won't work with earlier JDK versions (which was allowed for earlier Presto versions)?  That breaks backwards compatibility to some extent.  We should add a release note at least that the default configuration only works with JDK > 8u92

findepi:
@rschlussel fair question, but I don't have an answer. I would not bother with old Java updates, as they should not be used anyway. For example we had some (not committed) piece of code that compiled under u72 but did not work.

@rschlussel, @cawallin, are we OK with dropping support for `< u92` ?

cawallin:
It's an issue with upgrade though. I think we should work with at least the suggested configuration for 157t, as long as the recommended jdk version for 152 was 92+ I'm ok with dropping support for it, otherwise we should wait until next release.

findepi:
@cawallin  back to you

PR #335, opened 2017-03-14
Use 7070 port for presto server in tests

kokosing:
Use 7070 port for presto server in tests

While 8080 is not allowed on Teradata appliances, the 28384 is blocked
on Teradata EMR. 7070 should be fine for both cases.


kokosing:
I have run this job and it failed. I have no idea why, because it was passing just fine on my sandbox when I was using EMR cluster. While with port 28384 I was unable to locally make these tests to run successfully.

PR #336, opened 2017-03-16
Remove temp_dir with force

kokosing:
Remove temp_dir with force

For some (unknown reason) sometimes temp_dir is set to ''. Which
according to my understanding of the code should not happen.
It this happen then `rm -r ` (without file argument) is called which is
causing some error.


kokosing:
@rschlussel @cawallin I am sorry, but I am going to push it as it seems to be trivial change and I want to test emr jenkins build (it fixed locally the issue which was happening on emr).

Issue #348, opened 2017-11-13
Update docker-py to 2.6.0 from 1.5.0


kokosing:
When using current docker version and old docker-py it is possible to get some weird errors. I am getting this one:

```
"Cannot link to a non running container
```

[Here](https://stackoverflow.com/a/46849399/217852) someone suggest that we sholuld use `docker network` instead of link. 

Update would be nice, but it is hard to tell how much changes are needed to update this lib.

petroav:
Are you getting the error while running the product tests?

I feel like as long as the two versions are API compatible everything should work as expected. Did you try a Travis run with the new version?

kokosing:
Yes, I got this error while running product tests. I haven't yet started to do anything with the new version.

kokosing:
Fixed with #353

Issue #351, opened 2017-11-16
Cannot link to a non running container


kokosing:
https://api.travis-ci.org/jobs/303016970/log.txt?deansi=true

```

======================================================================
ERROR: test_connection_to_coord_lost (tests.product.test_server_install.TestServerInstall)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/prestodb/presto-admin/tests/product/test_server_install.py", line 384, in test_connection_to_coord_lost
    coordinator=down_node, pa_raise_error=False)
  File "/home/travis/build/prestodb/presto-admin/tests/product/standalone/presto_installer.py", line 52, in install
    rpm_name = self.copy_presto_rpm_to_master(cluster=cluster)
  File "/home/travis/build/prestodb/presto-admin/tests/product/standalone/presto_installer.py", line 98, in copy_presto_rpm_to_master
    if not self._check_rpm_already_uploaded(self.rpm_name, cluster):
  File "/home/travis/build/prestodb/presto-admin/tests/product/standalone/presto_installer.py", line 144, in _check_rpm_already_uploaded
    'ls ' + os.path.join(cluster.rpm_cache_dir, rpm_name)
  File "/home/travis/build/prestodb/presto-admin/tests/docker_cluster.py", line 292, in exec_cmd_on_host
    user=user)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/docker/utils/decorators.py", line 34, in wrapper
    return f(self, *args, **kwargs)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/docker/utils/decorators.py", line 19, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/docker/api/exec_api.py", line 71, in exec_create
    return self._result(res, True)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/docker/api/client.py", line 226, in _result
    self._raise_for_status(response)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/docker/api/client.py", line 222, in _raise_for_status
    raise create_api_error_from_http_exception(e)
  File "/home/travis/virtualenv/python2.7.12/lib/python2.7/site-packages/docker/errors.py", line 31, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation)
[...]

PR #352, opened 2017-12-08
Use docker networks and aliases instead links in tests

losipiuk:


tests/docker_cluster.py
```diff
@@ -178,6 +182,18 @@ def _create_host_mount_dirs(self):
...
+        self.client.networks.create(self._network_name)
+
+    def _get_network(self):
+        return self.client.networks.get(self._network_name)
+
+    def _remove_network(self):
+        try:
+            self._get_network().remove()
+        except NotFound:
+            pass
```
kokosing: Why is so? Maybe a warning?

tests/docker_cluster.py
```diff
@@ -206,8 +223,14 @@ def _create_container(self, image, container_name, hostname, cmd, **kwargs):
                                volumes={master_mount_dir: {'bind': self.mount_dir, 'mode': 'rw'}},
                                command=cmd,
                                mem_limit='2g',
+                               network=None,
+#                               network=self._network_name,
```
kokosing: ?

tests/docker_cluster.py
```diff
@@ -206,8 +223,14 @@ def _create_container(self, image, container_name, hostname, cmd, **kwargs):
...
                                **kwargs)
 
+        self._get_network().connect(
+            container_name,
+            aliases=[hostname.split('-')[0]])
```
kokosing: why do you use `split` here `-`? How `hostname` look like?

tests/docker_cluster.py
```diff
@@ -190,8 +206,9 @@ def _create_and_start_containers(self, master_image, slave_image=None, cmd=None,
             self.master,
             hostname=self.internal_master,
             cmd=cmd,
-            links=zip(self.slaves, self.slaves),
             **kwargs)
+
+        self._get_network().connect(self.master)
```
kokosing: master should be already connected, see line 204 where you call `_create_container`

tests/product/base_product_case.py
```diff
@@ -85,10 +85,7 @@ class BaseProductTestCase(BaseTestCase):
 query.max-memory-per-node=512MB
 query.max-memory=50GB\n"""
 
-    # The two strings below (down_node_connection_string and status_down_node_string) aggregate
-    # all possible error messages one might encounter when trying to perform an action when a
-    # node is not accessible. The variety in error messages comes from differences in the OS.
-    down_node_connection_string = r'(\nWarning: (\[%(host)s\] )?Low level ' \
+    network_unreachable = r'(\nWarning: (\[%(host)s\] )?Low level ' \
```
kokosing: Is this one is still valid once you ported tests to use network? I thought only one of these two errors should be expected.

kokosing:
Also make sure travis is passing. Other than that you did a great job!

kokosing:
Fixes #348 and #351

tests/docker_cluster.py
losipiuk: I think it is not need. Will remove.

PR #356, opened 2017-12-13
Fix tests to pass on RHEL7

anusudarsan:
Using the bash script to run presto-admin as non-root user were causing issues with RHEL7


anusudarsan:
closing this. not needed since the RHEL image itself was fixed

PR #358, opened 2017-12-14
Add travis step for building presto-admin artifacts

losipiuk:


bin/build-artifacts-in-docker.sh
```diff
@@ -0,0 +1,31 @@
...
+CONTAINER_NAME="presto-admin-build-$(date '+%s')"
+CONTAINER_DIR="/mnt/presto-admin"
+
+docker run --name ${CONTAINER_NAME} -v ${ROOT_DIR}:${CONTAINER_DIR} --rm -i ${BASE_IMAGE_NAME}:${BASE_IMAGE_TAG} \
+  env CONTAINER_DIR="${CONTAINER_DIR}" bash <<"EOF"
+    cd ${CONTAINER_DIR}
+    pip install --upgrade pip==9.0.1
+    pip install -r requirements.txt
+    pip install tox-travis
+    export PYTHONPATH=${PYTHONPATH}:$(pwd)
```
sopel39: is this external `PYTHONPATH` valid for docker container?

losipiuk: this is internal PYTOHNPAT

.travis.yml
```diff
@@ -32,7 +33,9 @@ before_script:
   - make presto-server-rpm.rpm
 script:
   - |
-    if [ -v SHORT_PRODUCT_TEST_GROUP ]; then
+    if [ -v ARTIFACTS ]; then
```
anusudarsan: looks like this has to go in a separate if-else, otherwise it fails at line 56

losipiuk: There was a type ARITFACTS vs ARTIFACTS. Just fixed that.

PR #369, opened 2018-02-02
Better port in use information

JerryLeooo:


JerryLeooo:
@kokosing Yeah I will check later when I have time

prestoadmin/server.py
```diff
@@ -511,10 +511,15 @@ def is_port_in_use(host):
     with settings(hide('warnings', 'stdout'), warn_only=True):
         output = run('netstat -ln |grep -E "\<%s\>" |grep LISTEN' % str(portnum))
     if output:
-        _LOGGER.info("Presto server port already in use. Skipping "
-                     "server start...")
-        error('Server failed to start on %s. Port %s already in use'
-              % (env.host, str(portnum)))
+        if 'command not found' in output:
```
findepi: if we `set -o pipefail` in `run(str)`, we should exercise better control... Or, perhaps, `netstat` should be run as command and `grep` as Py-level processing. (with `set -o pipefail` there is this problem with `grep` that it reports failure when no line matched, so you'd need `| { grep ... || true }`...)

@kokosing what do you think?

kokosing: Yes. That is much better!

PR #378, opened 2019-02-19
Added support for expanding environment variables in configuration files.

RootG:
Added support for expanding environment variables in configuration files.

For example:

Given an example node.properties files
```
node.environment=${ENV}
node.data-dir=/var/lib/presto/data
node.launcher-log-file=/var/log/presto/launcher.log
node.server-log-file=/var/log/presto/server.log
catalog.config-dir=/etc/presto/catalog
plugin.dir=/usr/lib/presto/lib/plugin
```
application can now resolve variable defined by `${ENV}` from environment variables.

- Definitions of a variable are `${variableName}` or `$variableName`. In addition to this, on Windows, `%variableName%` definition is supported as well.
- In case of an invalid or non existing variable definition, variable is left unchanged.


tests/unit/test_config.py
```diff
@@ -67,6 +67,14 @@ def test_get_properties(self):
                                 'd\\=': '4', 'e\\:': '5', 'f': '==6',
                                 'g': '= 7', 'h': ':8', 'i': '9', 'j': '${ENVIRONMENT_VARIABLE}'})
 
+    @patch('prestoadmin.config.os.path.expandvars', return_value='environment_variable_value')
+    def test_expending_environment_variable(self, environment_variables):
+        config_file = os.path.join(DIR, 'resources', 'environment_variable.properties')
+        conf = config.get_conf_from_properties_file(config_file)
+        self.assertTrue(environment_variables.called)
+        self.assertEqual({'a': 'environment_variable_value'}, conf)
```
kokosing, line 75: can we have an real test that is using actual environment variable, to see the configuration with expanded environment variables.

gorkemmulayim: Sorry for the late reply. I will work on this as well as automation failures.

kokosing:
@RootG Please ping once you done with your changes.

ddcprg:
this PR seems quite old. Is anyone going to revive it? @kokosing

